PR #1, opened 2016-03-09
Correct broken mysql_query (Pr/71 on old resources module)

PaulPoulain:
When I try to install this module, I am getting the error "Fatal error: Call to undefined function mysql_query() in /var/www/html/resources/install/install.php on line 64"

I suspect this was broken by 26ae17c which changes a lot of mysql_query calls, and also the commit message says "not yet tested"

### DIFF
--- a/resources/install/install.php
+++ b/resources/install/install.php
@@ -61,7 +61,7 @@ if ($step == "3"){
                         $sql_file = "protected/install.sql";
                         
                         // Assign the charset to specials chars
-                        mysql_query("SET NAMES 'utf8'");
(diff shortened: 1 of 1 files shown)


fondrenlibrary:
Can't duplicate the issue with the master branch.

remocrevo:
We need to remove this "mysql_query" function because the PHP mysql API is deprecated in PHP 5.5 (now the oldest supported version) and is fully removed from PHP 7.0 (released 5 months ago). For lots more details, see http://stackoverflow.com/questions/12859942/why-shouldnt-i-use-mysql-functions-in-php.

PR #4, opened 2016-03-10
platform creation bug in usage module (was #36 in non merged repo)

PaulPoulain:
fixes a bug that appeared in a November commit that creates a new Platform for every title.


benheet:
creating new platforms for every title sounds like a significant problem. is this problem more limited in scope than what it sounds? If not then this might need to be a priority pull request.

jeffnm:
I think we can close this request. It looks like the fixes from: https://github.com/ndlibersa/usage/pull/36/ have already made their way into the system in this commit https://github.com/Coral-erm/Coral/commit/aead8a896bd5c0ac0db08a4531f1f595ff7e358f#diff-fd07e4c6441b7582feb727729b3133e2

PR #6, opened 2016-03-11
Install admin accnt (was Auth12 on non merged repo)

PaulPoulain:
The main purpose of this pull request is to add a requested feature: specifying coral admin account during the auth install process.

The secondary (side-effect) purpose of this pull request is to cleanup the install script to make it easier to read and understand at a glance.

These changes work fine on my development workstation.


auth/install/index.php
```diff
@@ -109,94 +108,100 @@
...
+	} else if ($errmsg = $db->connect("$database->dbhost", "$database->dbuser", "$database->dbpass")) {
+		$errorMessage[] = $errmsg;
+
+	//next check that the database exists
+	} else if ($errmsg = $db->selectDB("$database->dbname")) {
+		$errorMessage[] = $errmsg;
+	} else if (!$db->query($create_admin_query)) {
+		$errorMessage[] = "Failed to create CORAL Admin";
+	//passed db host, name check, test that user can select from Auth database
+	} else if (!$db->query("SELECT loginID FROM $database->dbname.User WHERE loginID like '%$admin->coral_username%';")){
```
remocrevo: This line is hard to read. Could the variables be separated from the query string to improve readability? The result would be something like this:
...$db->query("SELECT loginID FROM " . $database->dbname . ".User WHERE loginID like '%" . $admin->coral_username . "%';")...

remocrevo:
Besides my one suggested change for code readability, these changes all look good from a code perspective.  But I haven't tested this code, and I think it deserves being tested again before merging it.  @jeffnm I think these improvements will help the proposed Combined Installer project.

jeffnm:
I've made some improvements to readability. The first was @remocrevo's suggestion to break out the variables in a query string. The other was to separate the DB username form and the CORAL admin username form visually. I did the same for the LDAP section. 

I ran two test installs. 
1. No LDAP and the default coral username and password.
2. LDAP and my LDAP credentials set as admin. 

Both succeeded and allowed me to login to the auth module with them.

PR #12, opened 2016-03-31
Database connection error in Resources

rkbrunn:
implements fix from May 11, 2015 in usage module allowing for use of single database connection when generating related resource lists. There was an error for too many database connections if the parent and child resources numbered over 140.


jeffnm:
The previous commit referenced is https://github.com/Coral-erm/Coral/commit/10976021c0900620daba19ef7381bdfd84dc233e

We've tested this on our dev server and I'm about to push it to production. It's a simple but needed bug fix.

PR #14, opened 2016-04-08
Licensing: Fix capitalization problem

remocrevo:
MySQL appears to be case insensitive, so capitalization doesn't matter in this case. BUT this is not true of every database platforms, so if we hope to support other platforms someday, this case needs to be fixed. Also, inconsistent capitalization makes the code confusing and may lead to bugs.


doug-hahn:
Comparing this fix with a fresh install / database, table creation the Column name is defined as qualifierID.  Testing appears to resolve the issue of inconsistent capitalization.

PR #18, opened 2016-04-11
Usage module (date issue)

rkbrunn:
Previously, users were unable to import data across multiple years (it would appear as if you could, but the data would overwrite the current year's data). This pull request fixes this error, allowing users to import data across multiple years in the past, and warns the user that data will be overwritten if using SUSHI on the uploadConfirmation.php page.

### DIFF
--- a/usage/admin/classes/domain/SushiService.php
+++ b/usage/admin/classes/domain/SushiService.php
@@ -657,76 +657,17 @@ class SushiService extends DatabaseObject {
 					// Create header for SUSHI file
 					///////////////////////////////////////////////////////
 					$header = $layoutColumns;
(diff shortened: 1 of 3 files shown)


jeffnm:
@rkbrunn I believe the `_()`  are needed for the translation tools to work.  Can you put them back in?

PaulPoulain:
Le 15/04/2016 16:49, Jeffrey Mudge a écrit :
>
> @rkbrunn <https://github.com/rkbrunn> I believe the |_()| are needed 
> for the translation tools to work. Can you put them back in?
>
>
right. _() are needed for translations

-- 
Paul Poulain, Associé-gérant / co-owner
BibLibre, Services en logiciels libres pour les bibliothèques
BibLibre, Open Source software and services for libraries

rkbrunn:
I didn't intentionally take them out. I'm thinking my IDE probably edited them out.

Will do.

rkbrunn:
Re-added the missing "_()" as requested.

PR #19, opened 2016-04-13
Various Bugs found during fresh install

doug-hahn:
Removed PHP shorthand usage
Fix bug in Org module
Fix bad merge commit

### DIFF
--- a/auth/admin.php
+++ b/auth/admin.php
@@ -61,7 +61,7 @@ if (isset($user) && ($user->isAdmin) && ($user->getOpenSession())){
 ?>
 </head>
 <body>
-<noscript><font face=arial><?= _("JavaScript must be enabled in order for you to use CORAL. However, it seems JavaScript is either disabled or not supported by your browser. To use CORAL, enable JavaScript by changing your browser options, then")." <a href=''>"._("try again")."</a>."?></font></noscript>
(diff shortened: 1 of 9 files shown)


jeffnm:
The `<?=` short code for `<?PHP echo` is not considered a short open tag in 5.4 or later. 
http://php.net/manual/en/function.echo.php says 

> Prior to PHP 5.4.0, this short syntax only works with the short_open_tag configuration setting enabled.

5.3 and prior are end of life, so I'm not sure that should matter.  In any case, `<?=` has been used for all translated strings in all modules. If we are going to change them it should be an intentional decision and applied to all the code. What do you think @PaulPoulain? 

Can you resubmit this with the bad merge fix and the org fix in two separate pull requests so they can be processed in while we discuss the `<?=` on the admin list?

doug-hahn:
Jeff, I don't see where <?= has been used for all translated strings in all modules.  Scanning the whole repo I find just 5 additional places where I think it should be removed.  For me it is a matter of consistency.  99% of the code doesn't use it.  So if a person is running a setup where short_open_tag is disabled they will see inconsistency in the code being executed.  

I'll break the stuff apart.

doug-hahn:
Closing Pull request and will re-submit.

jeffnm:
@doug-hahn I agree that we should be consistent. I tried to get the `<?=` changed to `<?php echo` statements previously and was told by @PaulPoulain and @remocrevo that we didn't need to do that because PHP 5.4 and later work just fine with short_open_tag disabled and older versions of PHP are end of life. I thought all the translation patches used the short code, but if they don't, then I don't know what to tell you. 

We definitely need some consensus on this and then stick with one way across modules.

PR #25, opened 2016-04-18
Remove PHP Shorthand to be consistent with rest of code

doug-hahn:


remocrevo:
For the record, here's the discussion where we concluded that the shorthand "<?=" was acceptable:
https://github.com/ndlibersa/auth/pull/3

But it does appear that the translation code in other modules (besides Auth) uses the full echo syntax.  I don't think we need to be strict about consistency in this case, especially if there are cases where the shorter syntax helps readability of the code.  But I generally agree that consistency is good.

benheet:
What was the action item for this from today's call? Was it to merge or no?

techsvcslib:
Merge.

Steve

jeffnm:
Great. I'll merge it.

PR #26, opened 2016-04-20
fix syntax error in mysqli_select_db in usage/install/install.php

bdewar:
Leaving out the $link param on line 145 makes the call fail, but the error message on
screen (from line 147) shows a blank MySQL message, since there was no error with the
$link itself.

Without this change, installing the usage module fails in PHP 5.5.9.


doug-hahn:
I verified this on a fresh install of the usage module.

PR #28, opened 2016-04-22
Bug Fixes and Enhancements for Organization and Resource Issues/Downtime Feature

jsavell:
### Enhancements
- Downtimes can be created without a Resolved Date/Time, and resolved in the future
- New Downtime form now uses Coral datepicker and selectable time for cross browser support
- Coral datepicker can now be used with AJAX content loaded by Thickbox modal

### Bugs Corrected

- Fix display bug when there is only one Downtime type
- Archived contacts were being included in list in New Issue form
- Fix bug that prevented Organization level issues from appearing in the Resource level Issue list
- Fix bug that prevented saving of updated Downtimes

There is also some general refactoring to cut down on code repetition and remove dead code.


resources/install/protected/update_1.4.sql
```diff
@@ -58,7 +58,7 @@ CREATE TABLE IF NOT EXISTS `Downtime` (
   `creatorID` varchar(80) NOT NULL,
   `dateCreated` datetime NOT NULL,
   `startDate` datetime NOT NULL,
-  `endDate` datetime NOT NULL,
+  `endDate` datetime DEFAULT NULL,
```
remocrevo: Because Resources 1.4 has just been released, which contained the initial Down Time features, this change needs to go in a separate upgrade script (presumably the 2.0 upgrade script, unless these changes are serious enough to release before then).

jsavell:
I added an incremented version (1.41) to the CORALInstaller.

I also added a mechanism in the installer for defining and detecting out of date versions when there are no new tables, only alterations.

resources/ajax_htmldata/getIssuesList.php
```diff
@@ -80,7 +80,7 @@ function generateIssueHTML($issue,$associatedEntities=null) {
 
 			if(count($orgIssues) > 0) {
 				foreach ($orgIssues as $issue) {
-					echo generateIssueHTML($issue,array(array("name"=>$orgData['organization'],"id"=>$organization->organizationID,"entityType"=>1)));
+					echo generateIssueHTML($issue,array(array("name"=>$orgData['organization'],"id"=>$organization->primaryKey,"entityType"=>1)));
```
remocrevo: Why this change? If organizationID doesn't produce the expected value, then there may be an underlying problem.

jsavell: That block of code was originally copied over from the Organizations module's version of the same Issues functionality, where the organizationID property returns the Org's ID.

The resources' version of the Organization class doesn't seem to do the same. I assumed it was just a module difference.

I'll look into it further.

jsavell: After more research, it's what I originally thought.

When the resources module is using the organizations module, the only safe way to access organization data is through the Resource class's getOrganization****() methods because the resource module's Organization class doesn't know about the other database or the additional table fields. 

This doesn't affect the getIssues method on the resources' Organization class, because all Issues live in the resource module's database.

Ideally, the resource module's Organization class would be able to model both the resource and organization module's Organization data, depending on Coral configuration.

I implemented a similar solution for the Issue class in the organization module:
https://github.com/Coral-erm/Coral/commit/a6312f18810efe2cb249aceba84b570eddf1ac0e

But that's beyond the scope of this PR.

jeffnm:
When entering an issue with no Problem Type the ajax fails with no user message. Error log shows that column downtimeTypeID cannot be null. This is not a blocker, but should be opened as a new issue if this PR is merged.

resources/install/protected/update_1.4.sql
jeffnm: Thanks for adding this to the 1.41 file. Could you remove it from update_1.4.sql?

Please also change resources/install/protected/install.sql to match. It currently has `'endDate' datetime NOT NULL`.

jeffnm:
I see that Downtime code has been added to the organizations module, but I'm not seeing it in the organization detail page. It should appear there under Issues, correct?

jsavell:
"I see that Downtime code has been added to the organizations module, but I'm not seeing it in the organization detail page. It should appear there under Issues, correct?"

Yes, Downtime gets loaded into the Issues tab by the same AJAX call ('getResourceIssueDetails') that loads the Issues, themselves:
https://github.com/Coral-erm/Coral/pull/28/files#diff-73a67b7c93343b10e91cb12e4e8e314f

If you're not seeing the Downtime header under Issues, something's not right.

organizationissuestab

resources/install/protected/update_1.4.sql
jsavell: Updated the install and 1.4 scripts.

jeffnm:
What I see is this: 
screen shot 2016-07-20 at 2 07 52 pm

I checked out the issue-feature-bugfixes branch, which is what this PR is based on.

jsavell:
Oh good, that's a solvable problem.

Set your Org module config to use the Resources Issues:
8. resourcesIssues=Y

The Org module already had a similar IssueLog feature. We didn't want to take it away from users who are already using it or aren't using the Resource module at all, so it's the one that's active by default.

jsavell:
Looks like this merge ( https://github.com/Coral-erm/Coral/commit/dadbad0d46dcc8160d9d9477e06cca1c30d85935#diff-ddd322bbfb9db7fc62387d3a62e49160 ) resulted in conflicts with this pull request.

I'll work on resolving them.

jeffnm:
That works! This looks pretty good to me. 

Thanks for addressing the conflicts. I'll merge when you are finished.

remocrevo:
@jsavell Did that config option make it into the install/update documentation?

jsavell:
The latest merge conflicts are resolved.

jsavell:
"Did that config option make it into the install/update documentation?"

Yes:
https://github.com/TAMULib/Coral-ERM/blob/issue-feature-bugfixes/organizations/install/upgrade.php#L125

<h3>Welcome to the CORAL Organizations upgrade for Version 1.3!</h3>
	This upgrade will connect to MySQL and run the CORAL Organizations structure changes. If you want to use the new Issues features in Resources module version 1.4, then you need to add the following line to the Organizations configuration file:
    <ul>
        <li>resourcesIssues=Y</li>
    </ul>
    Database changes for this version include:
	<ul>
		<li>Altering the IssueLog table to match the new IssueLog table in Resources v. 1.4</li>
		<li>Adding the IssueLogType table</li>
</ul>

tuxayo:
Hurray for refactoring and code cleanup!

@jsavell How hard was it to solve the conflicts? Is there a strategy to clean the files and limit the conflicts?

I had conflicts this morning also with this file because I edited it in two different PRs. And doing a rebase instead of a merge allowed to handle only the conflicts related with the cleanup without risking to loose the real changes.

jsavell:
@tuxayo 

Merging wasn't too bad, but I had decomposed some shared code between two methods in a couple of classes, and git wasn't sure what to make of it, so it took a little effort.

You may know this already, but be careful with rebase. It can easily ruin everyone else's day.
http://ryantablada.com/post/the-dangers-of-rebasing-a-branch

tuxayo:
> You may know this already, but be careful with rebase. It can easily ruin everyone else's day.
> http://ryantablada.com/post/the-dangers-of-rebasing-a-branch

Are you referring to issues with rewriting history of a shared branch? Or the example about a merge conflict issue (which I didn't managed to understand :-(  )

A note about conflicts, I quickly checked what would have been my conflicts if I had used merge instead of rebase yesterday. And I found at least one conflict which was more obscure and harder to solve due to the two edits (cleanup + actual changes) of the current branch being mixed instead of applied one by one.

PaulPoulain:
> You may know this already, but be careful with rebase. It can easily ruin everyone else's day.
> http://ryantablada.com/post/the-dangers-of-rebasing-a-branch
It CAN ruin everyone else day if you rebase on a branch that is not fully yours.
I can also be very useful for readability of the code history : if you submit a PR that has conflict and solve the conflict, once it's merged, it's hard to read. If you rebase your patch, and he's merged on top of HEAD, it has the same result, but/and the history is easier to read.
If both exist (rebase & merge), it's because both have their pros and cons.

jsavell:
As I said, it's something to be careful with. I didn't say it had no usefulness.

tuxayo:
@jsavell Is being careful to not use it on a shared branch enough or are there other pitfalls?

Issue #32, opened 2016-04-29
Ian issue


PaulPoulain:
see SC members for this issue, that has been declared privately for security reasons

PR #39, opened 2016-05-03
Exit after redirecting to login screen

sekjal:
Once it has been determined that a user needs to login before proceeding, stop executing PHP
code, in case the client does not respect Location headers by default (such as curl).

Resolves Issue #32 .

Applied to resources, licensing, organizations, management and usage modules.


remocrevo:
Thank you @sekjal for the patch, and for notifying us of the flaw.

Issue #17, opened 2016-04-11
Add enhanced cost history (new fields) to export


scottvieira:
New fields added to enhanced cost history need to be added to the export feature.  Note it appears dates for subscriptions are listed in export, but the dates are not actually exporting.

fondrenlibrary:
Solved

Issue #15, opened 2016-04-11
Add enhanced cost history (added fields) to print view 


scottvieira:
The new fields added to the enhanced cost history are not showing up in the print view.  These fields need to be added to the print view.  Also add notes fields to print view.

fondrenlibrary:
solved

PR #43, opened 2016-05-20
Pr4iss15.17: Add enhanced cost history new fields to export and print view

fondrenlibrary:
Used to solve issue #15 and #17 plus add basic date validity check to the date fields in cost history.


fondrenlibrary:
Will wait for Sirsi' s upcoming merge and then make corresponding updates to this PR.

resources/js/common.js
```diff
@@ -283,3 +282,40 @@ function postwith (to,p) {
   document.body.removeChild(myForm) ;
 }
 
+
+function isValidDate(dateString)
```
remocrevo: Instead of writing our own date validation code, we should consider using jQuery. Note: CORAL currently uses jQuery 1.4.4.  Here is an example:
http://stackoverflow.com/questions/6598386/jquery-validation-date

fondrenlibrary: Yes, it's really handy and better to use JQuery/Plugin, however at the time when this PR was being written, the version of jQuery  was found really behind.    Now, after pulling all latest commits into this PR, I decide to completely give up date validation because Subscription dates still look not required and can underlying mySQL can blindly set all invalid date inputs to   01/01/1970 after they are are submitted and can be changed at any time to good ones.

fondrenlibrary: After sync'd  with the master branch and  conflict resolving done, this PR can be tested again and ready to merge.  The following changes have been made to the original PR.
1. Date validation dropped.
2. Extra cost fields in the  Cost column in "exported" CSV dropped. If later better idea can be acquired about how to present all cost fields in the  Cost column, a new PR can be created.

tuxayo: @remocrevo 
> Note: CORAL currently uses jQuery 1.4.4.

Actually Coral uses [4 different versions of jQuery](https://github.com/Coral-erm/Coral/issues/84#issuecomment-236202203) so we have a broad choice ! ^^"

fondrenlibrary:
SC approved is assigned and ready to be merged (next week).

resources/js/common.js
fondrenlibrary: @remocrevo , nowadays it seems hard to find a working validation plugin for jquery 1.4.4 which is loaded by the Resource  pages.  On the other hand, the newest validation plugin can't handle leap years well as opposed to the validateDate() function added by this PR a while ago, though this is not a game changer at all.  Maybe we also need to discuss jQuery versions in the upcoming SC meeting. For now, as mentioned in my early comment, I completely drop out the date   validation function and we can re-add it later on separately. By the way, would you mind testing and merging this PR before this PR becomes conflicting again? Any direct editing and committing upon problems found is welcome. Thank you.

tuxayo: > By the way, would you mind testing and merging this PR before this PR becomes conflicting again?

Is there a test plan (instructions to setup and test) somewhere? So anyone could test it even without having followed this PR.

fondrenlibrary: @tuxayo , actually,  no particular test plan nor any integration test plan. The only test by hand plan is set enhancedCostHistory=Y and watch whether all the enhanced fields can  show up in Edit Cost Information and then watch whether Cost information can be saved. Also we need to observe whether those enhanced fields show in Print View which is under the Helpful Links. An extra benefit but not planned is this PR has fixed the error encountered upon saving Cost history. @tuxayo if your test proves fine to you, simply merge it. Thank you.

tuxayo: @fondrenlibrary 
>  actually, no particular test plan nor any integration test plan. The only test by hand plan is [...]

Maybe test plan weren't the right words but yes I meant manual instructions to be able to check that this PR works as intended without prior knowledge. So I (or others) could manually test it and contribute to build enough confidence so it could be merged quicker.

Thanks, the instructions that you provided are exactly what I needed. I hope to have the time to try them.

Issue #44, opened 2016-05-25
Add the final 1.x upgrade changes from old repos


jeffnm:
We need to apply the changes from the old standalone repos final 1.x releases so that we have the same base for the new upgrade processes. 

https://github.com/ndlibersa/

remocrevo:
@jeffnm Good call.  In particular, we need to make sure all the changes from the security patch are applied to this new repo.  And we may want to make a more thorough practice of always adding "exit" after any use of "header()".

jeffnm:
@remocrevo I think the security changes were already applied, but yes, we should make a habit of checking all use of "header()" to verify that they exit after.

remocrevo:
@jeffnm Actually, I think there are two or three cases where I added "exit" lines that didn't get applied to this repo. I'll try to find them...

https://github.com/ndlibersa/management/commit/e3ad3b6c2481306f6f6c93dec6bd70703a720724

https://github.com/ndlibersa/usage/commit/bbc25261c29d66616c1468f9f7ffe7994a7867a2

https://github.com/ndlibersa/licensing/commit/71b958ac461e7fb3f0a96ac676b85cf2b0db9828

Contrast those with the commit to this repo:

https://github.com/Coral-erm/Coral/commit/2e01ad35174d3af76af9f5d095811a52dfad13b3

PR #45, opened 2016-05-26
Update 1.X

baeiten:


jeffnm:
This should be resolving issue #44 

A few SQL statements that will be needed for 2.0 have been included in the update_next.sql files as well.

licensing/install/protected/update_NEXT.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `SFXProvider` CHANGE `SFXProvider` `sfxProvider` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT; 
+ALTER TABLE `expression` CHANGE `documentText` `documentText` TEXT CHARACTER SET utf8 COLLATE utf8_general_ci NULL;
```
remocrevo: The table name should be "Expression" (with capital "E").

Also note: In other upgrade scripts, we are using the command "ALTER TABLE [tablename] CONVERT TO CHARACTER SET utf8 COLLATE utf8_general_ci". We should confirm that these commands are equivalent.

remocrevo:
Thanks for pulling this together.

resources/install/protected/update_NEXT.sql
```diff
@@ -0,0 +1,79 @@
+DROP TABLE IF EXISTS `AuthorizedSite`;
+CREATE TABLE  `AuthorizedSite` (
+  `authorizedSiteID` int(11) NOT NULL auto_increment,
+  `shortName` varchar(45) default NULL,
+  PRIMARY KEY  (`authorizedSiteID`)
+) ENGINE=MyISAM AUTO_INCREMENT=1 DEFAULT CHARSET=utf8;
+
+DROP TABLE IF EXISTS `Resource`;
```
remocrevo: Wow, careful!  Why are we dropping and recreating the Resource table?  This will lose all data in the table.

jeffnm:
@baeiten Can you remove the update_next changes so this pull request only includes the changes brought over from the stand alone repositories? That way we can review your update_next.sql changes separately. 

As @remocrevo says, we never want to drop tables that have data in them already during the update process unless that data has already been migrated elsewhere.

PR #53, opened 2016-06-27
Resources: Fixes for modal size/placement on shorter viewports

jsavell:
The 'New Resource' form has a lot of inputs and the modal window wouldn't allow scrolling, which led to the submit/cancel buttons being unreachable on viewports with smaller vertical heights.

- Added media queries to optimize the placement of the modal based on the height of the client's viewport

- Removed the !important(s) that were preventing the vertical overflow property from functioning as intended.

> 
Using !important is bad practice and should be avoided because it makes debugging more difficult by breaking the natural cascading in your stylesheets.

https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

https://css-tricks.com/when-using-important-is-the-right-choice/

### DIFF
--- a/resources/css/thickbox.css
+++ b/resources/css/thickbox.css
@@ -57,7 +57,7 @@
 	display:none;
 	border: 4px solid #525252;
 	text-align:left;
-	top:50%;
+	top:40%;
 	left:50%;
 	width: auto !important;
 }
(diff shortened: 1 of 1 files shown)


bdroubay:
There are several modals that are too big or can grow to be too big to fit smaller displays.  My latest push includes CSS that overrides a ThickBox class (it appears at the bottom of the style.css for each module).  This allows the modal to scale based on the viewport.  If the window is resized, the modal will resize as well.  It is the best option we could come up with given the rush to get everything merged.  The functionality is great; however, the addition of the vertical scroll bar with the fixed width= passed in the ajax calls, causes a horizontal scrollbar to appear.  This could be suppressed with an overflow: none, but I think it needs to be applied to a div below the Thick Box class, and it may be necessary for the horizontal scroll to be available for smaller displays.

Thanks,

Brad

jsavell:
@bdroubay The vw/vh solution seems to work fine. I'd still like us to get away from !important, but looking at how often it's used elsewhere in the CSS, that's a bigger issue than the modal CSS.

PR #60, opened 2016-06-30
Use CSS to highlight links (rather than extra img & js)

jcuenod:
Lots of whitespace got automagically removed but the key is what I've removed from `common.js` and added to `indexstyles.css`. Side note: I'm working with css in a css/ folder in root (like the js folder). I think it's worth moving indexstyles into a css folder.


jeffnm:
The main menu hover effects with the CSS instead of javascript image swapping seems to work for me.

jcuenod:
Just realised that it doesn't handle a module not being installed properly. I'll submit another pull request. Sorry about that (thought I should start with something small though).

PR #61, opened 2016-06-30
Use CSS to highlight links (rather than extra img & js) - improved

jcuenod:
Lots of whitespace got automagically removed but the key is what I've removed from common.js and added to indexstyles.css. Side note: I'm working with css in a css/ folder in root (like the js folder). I think it's worth moving indexstyles into a css folder. Now also only highlights links that are to installed modules.


jcuenod:
I changed my mind about the text going darker so I've simplified the css in 9acbe0d. My git-fu is not so advanced that I can avoid commits that affect no files though (looking at you a10c2be).

jeffnm:
@fondrenlibrary Would you be willing to review this? It came from us, so I don't think I should be the one to merge it.

fondrenlibrary:
More than glad.

PR #62, opened 2016-07-01
add missing closing php tag

tuxayo:
For some reason the php 5.6 parser was fine with this.
But the PHP 7 parse was throwing a parse error


jeffnm:
Thanks @tuxayo

PR #70, opened 2016-07-06
update link to source code in footer 

tuxayo:
On thing to note is that I changed the link name from `GitHub Site` to `Coral is Free/Libre/Open Source` because when seeing GitHub site I though that Coral also had a [site hosted on GitHub](https://pages.github.com/).

If it's not appropriate just tell me and I'll change it.

### DIFF
--- a/licensing/templates/footer.php
+++ b/licensing/templates/footer.php
@@ -18,7 +18,7 @@
 ?>
 
 <br />
-<div class="footer">Copyright &copy; 2015. Licensing Module version 1.4<br/><a href="http://coral-erm.org/">CORAL Project Website</a> | <a href="https://github.com/ndlibersa/licensing">GitHub Site</a></div>
+<div class="footer">Copyright &copy; 2015. Licensing Module version 1.4<br/><a href="http://coral-erm.org/">CORAL Project Website</a> | <a href="https://github.com/Coral-erm/Coral/issues">Report an issue</a></div>
 </div>
 
 </td>
--- a/management/admin.php
+++ b/management/admin.php
@@ -36,7 +36,7 @@ if ($user->isAdmin()){
 <br /><span id='span_newUser' class='adminAddInput'><a href='ajax_forms.php?action=getAdminUserUpdateForm&height=202&width=288&modal=true' class='thickbox' id='expression'><?php echo _("add new user");?></a></span>
 <br /><br />
 <div id='div_User'>
-<img src = "images/circle.gif"><?php echo _("Loading...");?>
+<img src = "images/circle.gif" /><?php echo _("Loading...");?>
 </div>
 </td></tr>
 </table>
@@ -176,6 +176,7 @@ if ($config->settings->organizationsModule != 'Y'){ ?>
 <br />
 
 <script type="text/javascript" src="js/admin.js"></script>
+								</center>
 
 <?php
 }else{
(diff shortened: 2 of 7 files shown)


remocrevo:
Good idea. I suggest making it simply "CORAL is open source", or possibly changing the purpose of the link to "Report a bug" (linking to the CORAL Issues page on GitHub).

tuxayo:
Let's see what others prefer between the two.

veggiematts:
Free/Libre/Open Source seems a bit redundant at first, but Open Source alone is ambiguous.

So I personnally would go for Free/Libre/Open Source.

See: https://en.wikipedia.org/wiki/Free_and_open-source_software#FLOSS

I'm not for the "Report a bug" option, because it's not the only use of Coral's github.

jeffnm:
I understand why "GitHub Site" could be confusing to those who know about GitHub Pages.

I don't really like the Free/Libre/Open Source line though. 
I'd prefer just "GitHub" or "GitHub Repository". I'd also be ok removing the link and expecting the main website to provide access to the repository in appropriate ways.

tuxayo:
> I don't really like the Free/Libre/Open Source line though.

Fine, would "CORAL is open source" be okay?

>  I'd prefer just "GitHub" or "GitHub Repository". 

What's important is that it's the source code repo, not that it's on GitHub.

Disclaimer: I personally don't feel it's right to advertise GitHub as it's a proprietary service and it's becoming the center of the programming and open source universe. But I see that's making us loose time on wording issues. Sorry about that :-( , if GitHub gets more consensus then I have no problem using that so we can move on.

> I'd also be ok removing the link and expecting the main website to provide access to the repository in appropriate ways.

That's also a good solution.

Here is a pool: https://framadate.org/svpa94NjdWI1abHEe8qFtFOF/admin
We can add choices, and it's recommended to let a comment with the order of preferences so we have the **guarantee to be able to choose**

remocrevo:
Now that I think about it more, I realize this footer will only be seen by CORAL logged in users. They should already know that CORAL is open source.  Instead, what they need are things like helpful documentation (ask Web Committee where that should be) and a way to give feedback or report bugs (via GitHub Issues page).  We can provide those as separate links (which I prefer), or as @jeffnm said, we can send them to the CORAL website and let the website provide everything else.  Eventually, we could provide config variables such as a local tech support link which would be added to the footer, but that would take more work.

tuxayo:
The poll shows that `Report a bug` would be the best choice.

So I updated the message and link and **resubmit this PR for merging.**

I realize that I forgot to add an option about the "give feedback" part that was mentioned in the discussion. @remocrevo do you prefer `Report an issue`? That hopefully is close enough to `bug` (so that is not necessary to ask everyone again) and broadens a bit the scope to suggest that general feedback can be included.

PaulPoulain:
+1 for "issue" instead of "bug"

remocrevo:
+1 for "issue".

tuxayo:
So I updated the message and **resubmit this PR for merging**.

jeffnm:
Looks good. Could you update the management footer too? https://github.com/Coral-erm/Coral/blob/master/management/templates/footer.php

tuxayo:
@jeffnm Management footer is now up to date. I'm not sure for the version number, I took the [latest release on the old repo](https://github.com/ndlibersa/management/releases)

Here is how it looks
capture d ecran_2016-07-25_09-39-00

jeffnm:
Looks great! I'll merge now.

PR #86, opened 2016-07-19
Move session_start() out of class definition

jeffnm:
Invoking session_start() inside class CORALInstaller triggers an syntax error with an unexpected T_STRING. I'm moving it back up to the beginning of the file where Usage's CORALInstaller.php has it. 

Does any one know a reason why we shouldn't do this?

### DIFF
--- a/resources/install/CORALInstaller.php
+++ b/resources/install/CORALInstaller.php
@@ -1,4 +1,5 @@
 <?php
+session_start();
 require_once('../directory.php');
 
 if (!function_exists('debug')) {
@@ -8,7 +9,6 @@ if (!function_exists('debug')) {
 }
 
 class CORALInstaller {
-  session_start();
 
   public $db; // because CORALInstaller::query does unwanted things with result
   public $error;


jsavell:
I think it's fine there, but if we wanted to keep it in the class, session_start() could also be moved to be inside the __construct() method.

Also, session_start() got removed from that file, at one point, in favor of the CoralSession class: https://github.com/Coral-erm/Coral/commit/f734ace4df63adcf21ef01a6d0554061a18d88e1

Did it need to be added back in, or was that a merge side effect?

jeffnm:
Excellent question. It may not be necessary for that file anymore. It looks like the next commit inserted it again when the CoralSession class was being entered everywhere else. c3eef78

tuxayo:
So it should be completely removed if I understand correctly? Someone else in the mailing list got this bug again.
What is needed to be able to fix this so it's included in the next release? Just remove it and test that the Resource install works?

By the way, why doesn't this happens on every install?

remocrevo:
The syntax error might only happen for certain versions of PHP.  The user reporting it today was using PHP 5.6.24 (as part of Bitnami Wampstack-5.6.24-0).

jeffnm:
Actually, I think the reason some get it and others don't is that some people are downloading the master branch from https://github.com/Coral-erm/Coral which has the session_start() on line 11 while others are downloading the 1.4 release from the old repo (which is recommended) and as you can see, it does not have the session_start() in the class: https://github.com/ndlibersa/resources/blob/v1.4.0/install/CORALInstaller.php


As an aside, and this probably needs an issue to be opened, but I would like to suggest that we move to a development model where we only have release merges committed into master and everything else gets pulled into a development branch. That way when someone downloads the master they won't get unstable releases by accident. Something like this: http://nvie.com/posts/a-successful-git-branching-model/

jeffnm:
I'm going to go ahead and merge this. I haven't heard any reason why it would cause a problem, just a chance that we might not need it here at all. If we don't, we can remove it later, but for the sake of those who are downloading our master, I think it's important to get this fix into play.

PR #96, opened 2016-08-02
Small html, css, code style refactors and fixes

tuxayo:
See commit messages

### DIFF
--- a/auth/install/css/style.css
+++ b/auth/install/css/style.css
@@ -277,7 +277,7 @@ input.dp-applied {
 }
 
 .adminAddInput{margin-top:5px;}
-.licenseAddInput{margin-top:5px;margin-bottom;5px;}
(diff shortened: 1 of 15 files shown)


tuxayo:
Do not merge: https://github.com/Coral-erm/Coral/pull/96/commits/521f208b1bfd51f2f9e8f47c55f1d15a198678d5 (css: fix syntax errors) broke the footer on all modules except Licensing and Auth.

capture d ecran_2016-08-03_19-31-41
-
capture d ecran_2016-08-03_19-31-54
-
capture d ecran_2016-08-03_19-32-04
-
capture d ecran_2016-08-03_19-32-35

tuxayo:
What a pleasant surprise to discover that ` height; 120px;` did nothing as it was invalid so the browsers set the height themselves according their content size it seems.

So I removed this property which corresponds to the previous situation and shouldn't change anything in the style.

Issue #105, opened 2016-09-01
Resource list not showing up with MySQL 5.7


veggiematts:
The resource list is not showing up using MySQL 5.7 (PHP 7.0.8-0ubuntu0.16.04.2 and MySQL 5.7.13-0ubuntu0.16.04.2 in my case).

The following error message is found in the logs:

Uncaught Exception: Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'coral_resources_prod.R.titleText' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by

PR #106, opened 2016-09-01
Issue 105: Resource list not showing up with MySQL 5.7

veggiematts:
This patch fixes resource list for MySQL 5.7.

See http://stackoverflow.com/questions/34115174/i-am-getting-an-error-in-mysql-related-to-only-full-group-by-when-executing-a-qu for reference.

Fixes issue #105

### DIFF
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -1366,7 +1366,7 @@ class Resource extends DatabaseObject {
 		$alphArray = array();
 		$result = $this->db->query("SELECT DISTINCT UPPER(SUBSTR(TRIM(LEADING 'The ' FROM titleText),1,1)) letter, COUNT(SUBSTR(TRIM(LEADING 'The ' FROM titleText),1,1)) letter_count
 								FROM Resource R
-								GROUP BY SUBSTR(TRIM(LEADING 'The ' FROM titleText),1,1)
+								GROUP BY SUBSTR(TRIM(LEADING 'The ' FROM titleText),1,1), titleText
 								ORDER BY 1;");
 
 		while ($row = $result->fetch_assoc()) {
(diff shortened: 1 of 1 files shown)


tuxayo:
Also works with `MariaDB 10.1.16` + `PHP 7.0.10` + Arch Linux
Tested to check if it doesn't break compatibility.

tuxayo:
No issues with this patch when doing three installs on different machines with MySQL5.7 + Ubuntu 16.04 + PHP 7

veggiematts:
Can someone have a second look on this? It's a really small patch :)

resources/admin/classes/domain/Resource.php
```diff
@@ -1366,7 +1366,7 @@ public function getAlphabeticalList() {
 		$alphArray = array();
 		$result = $this->db->query("SELECT DISTINCT UPPER(SUBSTR(TRIM(LEADING 'The ' FROM titleText),1,1)) letter, COUNT(SUBSTR(TRIM(LEADING 'The ' FROM titleText),1,1)) letter_count
 								FROM Resource R
-								GROUP BY SUBSTR(TRIM(LEADING 'The ' FROM titleText),1,1)
+								GROUP BY SUBSTR(TRIM(LEADING 'The ' FROM titleText),1,1), titleText
```
remocrevo: I think we actually want the GROUP BY clause to match the first part of the SELECT clause. I'm guessing that would prevent the warning. I would suggest testing this:

GROUP BY UPPER(SUBSTR(TRIM(LEADING 'The ' FROM titleText),1,1))

veggiematts: Your solution worked on my instance.

I would go for any solution, but I'm not sure which one is the best, performance-wise (or if they are equivalent)

benheet: Looks like a decision needs to be made here between two different solutions. @remocrevo or @veggiematts do either of you have the time to modify this to use the solution that prevents the warning? If not would it be acceptable to accept this as is and then add an issue for making this modification later?

benheet:
Let's go with Remington's proposed solution on this. @veggiematts can you update this to use Remington's proposed solution which you successfully tested and then we'll merge it in?

veggiematts:
Done.

Issue #110, opened 2016-09-05
Reports module: install issue with MySQL 5.7


tuxayo:
At the step "MySQL info with permissions to create tables and Link Resolver info"

When submitting I get:

```
All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead

For statement:
CREATE TABLE IF NOT EXISTS `ReportParameterMap` (
    `reportID` int(11) default NULL, `reportParameterID` int(11) NOT NULL auto_increment,
    `parentReportParameterID` int(11) default NULL,
    PRIMARY KEY (`reportID`,`reportParameterID`) 
    )
ENGINE=MyISAM AUTO_INCREMENT=0 DEFAULT CHARSET=utf8
```

PR #112, opened 2016-09-05
Fix default timestamps to be compatible with MySQL 5.7 - issue #50

### LINKED ISSUE #50: Invalid default date value
When trying to install the latest version of Coral resources, I get this error:

> Invalid default value for 'updateDate'
> 
> For statement: CREATE TABLE `ResourceNote` ( `resourceNoteID` int(11) NOT NULL auto_increment, `resourceID` int(11) default NULL, `noteTypeID` int(11) default NULL, `tabName` varchar(45) default NULL, `updateDate` timestamp NOT NULL default '0000-00-00 00:00:00', `updateLoginID` varchar(45) default NULL, `noteText` text, PRIMARY KEY (`resourceNoteID`) ) ENGINE=MyISAM AUTO_INCREMENT=1 DEFAULT CHARSET=utf8`

I am using PHP 5.6 and MySQL 5.7

For newer verisons of MySQL, strict mode has been changed and you cannot use '0000-00-00' as a default date without changing some settings in MySQL.

My suggestion instead would be to instead set the default value to be "NOW()", or to enable NULL values.

http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-changes
http://stackoverflow.com/questions/25349126/how-can-i-set-the-default-value-of-a-field-as-0000-00-00-000000

Happy to make a PR for this if it sounds like a reasonable solution, however I'm not entirely sure yet how that field is used or what makes most sense for existing installs, or how this would be handled in an upgrade.

tuxayo:
An install was successfully done (except for module Reports #110 ) on MySQL 5.7, PHP 7, Ubuntu 16.04.

And a resource was created.

Fixes #50

### DIFF
--- a/auth/install/create_tables_data.sql
+++ b/auth/install/create_tables_data.sql
@@ -1,7 +1,7 @@
 CREATE TABLE `Session` (
   `sessionID` varchar(100) NOT NULL default '',
   `loginID` varchar(50) default NULL,
-  `timestamp` timestamp NOT NULL default '0000-00-00 00:00:00',
+  `timestamp` timestamp NOT NULL default CURRENT_TIMESTAMP,
   PRIMARY KEY  (`sessionID`)
 ) ENGINE=MyISAM DEFAULT CHARSET=latin1;
 
--- a/licensing/install/protected/install.sql
+++ b/licensing/install/protected/install.sql
@@ -48,7 +48,7 @@ CREATE TABLE IF NOT EXISTS `Expression` (
   `expressionTypeID` int(10) unsigned NOT NULL,
   `documentText` text,
   `simplifiedText` text,
-  `lastUpdateDate` timestamp NOT NULL default '0000-00-00 00:00:00',
+  `lastUpdateDate` timestamp NOT NULL default CURRENT_TIMESTAMP,
   `productionUseInd` tinyint(1) NOT NULL default '0',
   PRIMARY KEY  (`expressionID`)
 ) ENGINE=MyISAM AUTO_INCREMENT=1 DEFAULT CHARSET=utf8;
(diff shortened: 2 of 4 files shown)


jamesrf:
Thanks for doing this, but I think this fix should hit this script as well to give people an upgrade path:
https://github.com/Coral-erm/Coral/blob/master/resources/install/protected/update_NEXT.sql

But this would also require updating all the old dates in the existing table that are still 0000-00-00.  But I don't know what impact this might have on an existing data set?

tuxayo:
Ow I forgot about existing data!

Unfortunately as I didn't have the time to do this during my internship so I won't be able to finish this work.

tuxayo:
Another thing, it seems that an install that upgrades to MySQL 5.7 with the default date still being 0000-00-00 on the table and existing data with that date still works for some reason.

Can you confirm @veggiematts ? IIRC you have upgraded to MySQL 5.7 on your machine without needing this patch.

Anyway data + schema migration is still needed for consistency.

benheet:
It looks like this had a major issue as it didn't deal with existing data. That at least to me suggests this pull request should either be modified by someone or closed rather than leaving it open. thoughts?

cjc66:
This is Brad on Carla's computer. If the dates are already in the table, then there is not a problem with them.  The problem is that 5.7 won't accept 0000-00-00 00:00:00 as a valid date, which prevents the record from loading.  I don't think existing data is a concern.

benheet:
The decision of merge vs close on this pull request is still beyond me as
the issue isn't one I can confidently understand. Jeff or Remington can you
share your thoughts on this given the latest from Brad?

-
Ben

jeffnm:
Looking at the files that have been changed in this request, I see no reason not to merge the changes. 

Brad is correct, existing data is fine. However, since these SQL statements are modifying the table definition for the default values for new data from `0000-00-00 00:00:00` to the `CURRENT_TIMESTAMP`, I think we do need to `ALTER` existing tables to match, otherwise an existing installation will develop issues when migrating from an older version of MySQL and then attempting to new data. 

Let me leave this open a little longer, and I'll get the `ALTER` statements into the 2.0.0 upgrade SQL.   I'll assign myself to this ticket to come back and merge it in.

bdroubay:
The problem will only occur with a database upgrade. These are two different issues. The pull request is good, as it won't affect former versions of MySQL. Upgrading a database is a different story. There can be a lot of challenges there. Liquibase would handle a lot of them, but it is an issue that needs discussion.

I hope that gave so clarification.

Thanks,

Brad

jeffnm:
I've added these modifications to the 2.0.0 upgrader SQL files. Since there are no objections to this pull request itself, I'll merge it now.

Issue #114, opened 2016-09-06
Choose a testing framework to begin building an automated test suite


tuxayo:
# TL;DR

We need tests in Coral. GUI tests are the most efficient way to **start** testing. Codeception is the best option for GUI testing.
# Motivations: why do we need a test suite? (skip if already convinced)
- detect regressions on the most critical use cases and features to ensure that they never break.
- detect regressions earlier to allow the developer that already has the context to fix them immediately. Instead of someone else much later who will spend more time to understand what went wrong.
- spend less time manually testing the most common use cases.
  - important when some of the features require dozens of clicks and more than 20 sec to be tested (setting up the DB in a particular state) which is tedious, error prone and can't be done after each change.
  - even more important considering the amount of refactoring work that must be done and that the lowest layers can impact all the features.
  - as the test suite will become more comprehensive the QA work before each release will stay constant or will decrease (depending on the test coverage). While without automated tests, it will increase thus making harder to release as time goes on.
- less risk of breaking something else when doing a hotfix.
- document the most common and some complex use cases for the developers. Useful when not knowing well the domain (ERMS) and Coral. It allows to integrate new devs easier and reassure them about not breaking everything without noticing it.
- easily test with multiple versions of PHP, MariaDB/MySQL and reduce  the cost of supporting a new DBMS. Over the years it will avoid some bad surprises (`mysql` API dropped in PHP 7, [default date values changes](https://github.com/Coral-erm/Coral/issues/50) in MariaDB/MySQL)
- make comprehensive and reproducible performance benchmarks
- as Coral will grow in features, complexity and users, the benefits of having a more and more comprehensive test suite will increase. It's already risky to touch the lower layers without breaking something far away from there that will only (hopefully) be detected in the QA phase before the next release (mysql→mysqli migration). Thus having to test everything manually is a barrier to refactoring. Refactoring is essential on the long run to improve the code quality and prevent the codebase to become "untouchable".
### database merge (between all modules) #65

It could potentially introduce regressions in every part of the app. If started now, this work would require a huge amount of manual testing of the main features at many stages of the refactoring process. Having them automated would allow to save a lot of time + fix regressions just after they were introduced instead of hours/days latter.
### deduplicate common code

For example in `*/admin/classes/` for each module we have many identical(or almost) classes.

Same problems as for merging the DB
### separate view from logic #78

[...]

PR #115, opened 2016-09-06
Foundations of test suite

tuxayo:
See #114

Section "Codeception + real browser (PhantomJS)"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Composer: installed libs folder
 vendor/
 
+# Composer: installed binaries: PhantomJS, Codeception
+bin/
+
 # Codeception temp and generated files
 tests/_output/
 tests/_support/_generated/
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,6 @@
+# Composer: installed libs folder
+vendor/
+
+# Codeception temp and generated files
+tests/_output/
+tests/_support/_generated/
--- a/README
+++ b/README
@@ -2,6 +2,9 @@
 Installation
 These files are intended to be copied into the /coral/ directory on your web server.  The index should only display active links for the modules you have under the directory as long as the name matches the naming in Github (e.g. /coral/resources/)
 
+**************************
+Automated tests
+See the file TEST.md
 
 **************************
 License / Copyright notices
--- a/TESTS.md
+++ b/TESTS.md
@@ -1,0 +1,71 @@
+# Setup
+## Install composer
+### Create /bin folder
+`mkdir bin`
+### Download composer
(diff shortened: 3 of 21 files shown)


tuxayo:
Wait, `/tests` shouldn't be accessible from the webserver. Does putting a .htaccess with [this content](https://github.com/Coral-erm/Coral/blob/master/management/admin/.htaccess) is enough?

remocrevo:
@tuxayo Thanks for your work on this!  Is it common to include the composer.lock file in a code repository, or is that dependent on a local server's configuration?

tuxayo:
@remocrevo Yes composer.lock is necessary as it freezes the versions of the dependencies. To ensure we all use for example PhantomJS `2.1.1` and not some random `2.X.Y` version as the composer.json only specifies `> 2.0.0`

It avoids issues due to different versions without having to manually set one exact version in composer.json

https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file

jeffnm:
The framework seems to be working. I was able to install the pieces and get things set up, but when I run the `bin/codecept` command as described in TESTS.md the `ResourcesCept` test fails. Looking at the generated screenshot and logs it seems that it is trying to enter the form information before the AJAX form loads. 

This is the screenshot of the step 2 (`Fill field "#titleText","test resource"`) failure. If I do it manually I get the form to load properly - that bar only flashes on the screen momentarily. 

resourcescept fail

I just ran it with php 5.5 and 7.0 on localhost.

tuxayo:
@jeffnm I've never encountered this on the 4 machine where I ran the test a thousand times in a row. What a bad surprise. I added a step to ensure that the form loads, can you try again?

jeffnm:
@tuxayo Thanks for all your work on this!  I was testing on a fresh install, so once I created a ResourceType your tests passed all assertions. I tested php 5.5 and php 7.0 and this test ran and passed all assertions on both. 

I did check that my lack of a Resource Type was not the problem with the previous commit, but that commit still fails with the incomplete modal screenshot, even after adding a Resource Type. This latest commit works though, so that's great!

tuxayo:
> I did check that my lack of a Resource Type was not the problem with the previous commit, but that commit still fails with the incomplete modal screenshot, even after adding a Resource Type.

Thanks for double checking this. We now have the confirmation that there can be some issues related to loading times that happen on one machine but never happen on others. That will cause bad surprises but the fix will be simply adding a generic `$I->waitForPageToBeReady();` or a specific `waitForElement` or `waitForText`

tuxayo:
Surpersesded by #118

PR #118, opened 2016-09-16
Foundations of test suite with DB isolation

tuxayo:
This PR supersedes #115.

Follow [TEST.md](https://github.com/tuxayo/Coral/blob/isolate-db-in-tests/TESTS.md) to try it.

It adds:
### A separated database for tests
Before each test, it will load a dump of a clean install with [few things already set up.](https://github.com/tuxayo/Coral/blob/isolate-db-in-tests/tests/_data/test_database.sql#L10)
Which will avoid issues of a test leaving garbage that will disturb latter tests.
Test don't need anymore to clean themselves.
Your existing data won't interfere.

### The login is moved to it's own test
Which allowed to fix other issues that would immediately arise when another test would have been added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,6 @@
+# Composer: installed libs folder
+vendor/
+
+# Codeception temp and generated files
+tests/_output/
+tests/_support/_generated/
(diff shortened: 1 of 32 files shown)


jeffnm:
We're pretty happy with this in our tests here. We have run into a few configuration issues that we're trying to iron out and document, so I am merging this into a test_suite branch so we can make those changes before merging into master.

PR #127, opened 2016-10-18
Unified installer

jeffnm:
This is the unified installer. 

The upgrader needs some additional testing before it's ready for a pull request.


fondrenlibrary:
There is a conflict, can it be solved in quick way by the author this PR?

jeffnm:
It looks like it's just the index.php file, so I think we should be able to fix that without too much trouble.

jeffnm:
We're going to try to merge the updater into this pull request because it uses a lot of the same code.

jeffnm:
Right now I believe this pull request includes everything for the installer and the updater, including the few SQL updates that are needed from 1.x to 2.0.0  

I'd welcome further testing and feedback.

fondrenlibrary:
@jeffnm we will test it  as an installer and as an updater respectively next week.

jeffnm:
@fondrenlibrary has identified an issue with the upgrader when a config file is missing but the person installing the upgrade indicates the module is actually installed. We're working on a fix.

jeffnm:
For those wanting to test this code:

# Clean Installation Instructions

1. Clone this pull request
2. Put the directory into a web server directory
3. Navigate to the root folder in browser (main CORAL index page)
4. Follow the Installer prompts to install the desired modules


# Upgrade Instructions
1. Make sure you have the latest version of the 1.x modules installed.
2. Make the usual directory and database backups (so you can restore/recover if something goes wrong)
3. Copy the new code over the main CORAL directory 
4. Visit the url for the main coral index page
5. You should get a warning that CORAL can’t tell if it is already installed yet or not and it will offer you an option to install it if not, or to say that it’s already installed and that you are attempting to upgrade/repair it.
6. Pick the option that says it is already installed (upgrade or repair).
7. Pick the modules that you have installed. At this point, it will then begin running through the dependencies and requirements, including checking your permissions. It will run any SQL changes that we have included in the modules upgraders. 
8. At the end you should have a working 2.0.0 installation. It should be very similar to the clean install process.

jeffnm:
The bug which caused missing configuration files to corrupt the installation has been resolved now. The installer will no longer allow you to upgrade a module that is missing its configuration file.

jeffnm:
I'm going to merge this into master now, per the discussion on the Steering Committee list. 

Any additional bug fixes that are needed will come in under their own pull requests.

PR #130, opened 2016-10-24
Test suite

jeffnm:
The additional tests and a few more acceptance tests have been created on this branch. There are definitely further improvements to be made, especially since we only have one acceptance test per module so far, but I think this is enough progress to merge into the master branch. 

I'd like to see us open some issues for additional tests that could be written.


benheet:
Jeff submitted this a while ago. I propose we skip the 2nd party review and allow Jeff to merge the pull request.

jeffnm:
It's been two weeks since Ben's suggestion. Since these tests don't change any core code I think it's safe to skip additional review at this point, although I'd love for others to start contributing tests as well. 

**I'll merge it tomorrow if I don't hear any objections.**

remocrevo:
@jeffnm I agree that no additional review is needed right now.

fondrenlibrary:
I Just got a little bit  time to test it, though a very late comment, the testsuite requires PHP5.6 based on the output of my running Composer Installer. This requirement is  too high, given that many sites are still running Redhat (RH) 6 and under let alone RH 7 which only offers php55 as its highest available PHP version.

benheet:
Is the suggestion here then that the merge should be undone or that we just need to provide this piece of clarification within the documentation concerning PHP version for using this component?

fondrenlibrary:
Because it will affect  only developing and testing side not deployment, I think we should keep the merge but may in the README add some explicit caveat for developers saying PHP5.6+ is only needed specifically for the test suit, though composer will indicate this dependency issue in its runtime as well.

PR #117, opened 2016-09-13
Workflow enhancements: restart, archive and edit workflow

veggiematts:
**New features:**
- Restart a completed workflow: Users shall be provided the ability to restart a new workflow by
  selecting a workflow from a pre-defined workflow list or restart the current completed workflow.

workflow_1
workflow_2
- Provide ability to archive previous workflows: Previously completed workflows contain valuable
  information, such as date when tasks are completed, and such information shall be kept.

workflow_3
- Provide the ability to add a step in the workflow for a particular resource: The ability to add a step
  offers users greater flexibility to determine a customized workflow for a specific resource. CORAL
  provides a globalized add a step function in the Admin section, which applies to all resources
  associated with this workflow, not on individual resource level.

workflow_4
workflow_5
workflow_6

PR #131, opened 2016-10-25
Email reminder on workflow steps

veggiematts:
Note: this PR is based on PR #117 and therefore contains all of its commits.

This patch allows to trigger an email reminder after a resource step is assigned to a user.
The delay (day-based) can be configured for every resource step:

mail_reminder


benheet:
@veggiematts can you say more about what the reminder delay is doing? Is it sending a reminder email to the assigned person x number of days after the workflow step is activated? I don't think the workflow template is the proper place to set email reminders that go to a specific person or group. I'd enable a workflow reminder as a global feature that went to anyone assigned to any active workflow step X days after the step was active (if not yet complete) OR i'd leave it to the individual to select whether they wanted reminders somewhere within the My Queue page. This implementation of a reminder feature seems more complex then it needs to be and I'd rather not see this merged without further discussion.

veggiematts:
It sends a reminder email to the assigned persons (each member of the group assigned to this step) n days after the step has been started, where n is the reminder delay.

I thought the specification was discussed with the community prior to this implementation (in the coral-admin mailing list?)

Maybe @xsong9 can tell us more about what was discussed?

xsong9:
@benheet The email reminder can be enabled in the Edit the Workflow window in the Routing tab associated to a resource. It's part of the local workflow enhancement implementation #117 . Setting a workflow reminder as a global feature will not work because the number of delay days for a step varies depending on a specific resource. For example I have a workflow setup for ebook package acquisition and it includes three steps: licensing, invoicing and setup access. I can use the same workflow for both backfile purchase and frontfile 2017 purchase. The backfile access can be set up right after the purchase, but the frontfile access need to wait until the titles are available, which means a couple of months later or even longer.  With this implementation, I can set up different dates for the access step for the backfile and the frontfile locally. It provides flexibility and can avoid creating separate workflows in the Admin.  

Please let me know if you have more concern or question about this feature.

benheet:
Note this PR contains the commits from #117 but I have reservations about the additional alert features added in this PR. I propose that we return to and merge PR #117 and deal with the additional functionality of this PR later.

benheet:
@veggiematts @xsong9 this pull request has been approved but it has conflicts that need to be resolved before it's merged. Can you take responsibility for resolving the conflicts please?

jeffnm:
I've added the 2.0 beta milestone. If we can get the conflicts resolved in time, we can include it. If not, I'll remove the milestone.

veggiematts:
Conflicts resolved on forked branch, but it somehow doesn't seem to be updated in this PR.

veggiematts:
Don't merge into master yet, there are still issues in this branch.

veggiematts:
Nevermind, the problem was on my coral instance. As far as I'm concerned, we're good to go!

jeffnm:
There aren't any conflicts any more, but I'm seeing a bunch of file changes that should already be in coral-erm/master. For example, all the of the Test Suite files are showing up in this pull request. I guess that's not a problem since the end result will be the same, but it strikes me as odd.

veggiematts:
@jeffnm Do you know what could have caused this?

jeffnm:
@veggiematts I don't. Perhaps something in the way the rebase was done? 

Do you think it's a problem?

jeffnm:
@veggiematts beautiful. I'll merge it.

PR #132, opened 2016-10-28
Improvements to the resource import functionality

mjlassila:
This pull request includes  few extra fields to the resource import functionality (including user limit count, subscription start and end dates, order number and purchasing site). It also changes the default import encoding to UTF-8.

### DIFF
--- a/resources/admin/classes/domain/PurchaseSite.php
+++ b/resources/admin/classes/domain/PurchaseSite.php
@@ -35,6 +35,18 @@ class PurchaseSite extends DatabaseObject {
 
 	}
 
+	public function alreadyExists($shortName) {
+		$query = "SELECT count(*) sitecount FROM PurchaseSite WHERE UPPER(shortName) = '" . str_replace("'", "''", strtoupper($shortName)) . "';";
+		$result = $this->db->processQuery($query, 'assoc');
+		return $result['sitecount'];
+  	}
+
+  	public function getPurchaseSiteIDByName($shortName) {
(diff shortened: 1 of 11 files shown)


veggiematts:
Can you explain this point:
"Function for reading UTF-8 encoded files." ?

The import tool already supports utf-8 files. It seems to me that your utf8_fopen_read function converts windows-1250 encoded files to utf-8 without testing the encoding first. This will mess up true utf-8 files.

Or am I missing something here?

mjlassila:
Oops, my mistake. Change to fopen()-function is not needed. At first I didn't notice that our data had broken Unicode but when I cleaned it using [ftfy](https://ftfy.readthedocs.io/), import works fine. Sorry about the mess.

mjlassila:
It seems that mb_convert_encoding() is giving me problems at least in our server environment when working with files which are  properly UTF-8 from the origin.  Strings such as `Revue d'histoire des mathématiques` are being mangled to `Revue d'histoire des mathÃ©matiques`. When I remove the call to mb_convert_encoding() characters are being imported correctly. Could it be that the problem lies in our server environment and not in Coral, even though in our case, bypassing mb_convert_encoding() fixes the problem?

benheet:
This is the only open pull request for the import feature, correct? Is the conflict in the pull request something that can be addressed easily?

benheet:
@veggiematts reading an earlier comment it was proposed that perhaps this was not a problem with coral but rather a local problem in the server environment. Is that your thought on this too, in which case we should close the pull request.

mjlassila:
Yeah, I'll close. I'll fix conflicts with current master later and make a new pull request including only the changes related to the improvements to the resource import functionality.

PR #134, opened 2016-11-01
Fixing poorly formatted NULL acquisitions dates

jeffnm:
The acquisition currentStartDate and currentEndDate were being passed `'null'` which mysql normally rejects as incorrect SQL for a `DATE` field. All the other `NULL` values being passed into the update query were coming in correctly as `NULL`, so I changed this file to match which fixed our database error.


benheet:
Can we skip the 2nd party review step on this and allow Jeff to merge?

PR #135, opened 2016-11-01
Ridding "Notice: Undefined index:"

librariman:
I repeatedly got the error "Notice: Undefined index: HTTPS in C:\xampp\htdocs\coral\resources\admin\classes\common\Utility.php".  

I should note that regardless of what I use for the URL's prefix (www, http:, https, etc), it always removes it.  so **www.mycoral.com** becomes **mycoral.com**.

### DIFF
--- a/resources/admin/classes/common/Utility.php
+++ b/resources/admin/classes/common/Utility.php
@@ -68,7 +68,7 @@ class Utility {
 	//returns page URL up to /coral/
 	public function getCORALURL(){
 		$pageURL = 'http';
-		if ($_SERVER["HTTPS"] == "on") {$pageURL .= "s";}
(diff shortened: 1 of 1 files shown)


jeffnm:
@librariman Thank you for your contribution! 

Just to clarify, the removing of `https://`, `http://`, or sub-domains like `www.` will be handled by apache.  Apache is probably configured to send all traffic to `mycoral.com` on port 80 (which is `http://` which most browsers won't bother displaying) and that's what's causing the effect you see there. 

We have a lot of undefined index notices in the CORAL codebase, so I'm always happy to see any of them addressed. They don't usually break anything, but they do make digging through error logs a lot more difficult.

benheet:
@jeffnm with your initial review of this was there any concern in merging?

jeffnm:
Thanks @benheet for the reminder. It was a straightforward change. I've merged it now.

PR #149, opened 2016-12-20
Fix duplicate Parent bug in updateProduct

### LINKED ISSUE #148: Adding a new Parent to a resource results in two instances of that Parent
When adding a new parent to a resource through the Edit Resource form, the parent is associated with the resource twice.

jsavell:
Resolves #148 

This is a quick fix to resolve the bug.

A more thorough refactor of the javascript and HTML is needed, since it seems to be halfway between an original, single parent version, and the current multiple add version. 

This makes it difficult to debug, and performance could be improved with more precise jQuery selectors and HTML dedicated to adding multiple parents in one edit.

### DIFF
--- a/resources/ajax_forms/getUpdateProductForm.php
+++ b/resources/ajax_forms/getUpdateProductForm.php
@@ -120,7 +120,7 @@
 
            <span id="newParent">
            <div class="oneParent">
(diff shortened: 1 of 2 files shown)


jeffnm:
This looks like a really straightforward bug fix, I don't see how it could cause other issues, so I'd like to go ahead and merge this.

PR #152, opened 2017-01-18
Import tool: Preview the mappings

veggiematts:
preview_1
Columns numbers and their titles are displayed when creating or editing the import profile.

preview_2
Before the real import, a preview of the imported values is displayed, as well as the column to coral field mapping. The number of elements (resources, parents, organizations...) that will be created is also displayed.


jeffnm:
Reopened on development branch at #171

PR #154, opened 2017-01-18
Import tool: new fields

veggiematts:
The following fields are added to the import tool: acquisition type, fund code, cost, order type and currency.

add_1 
add_2

Acquisitions lines are created automatically if fund code and cost are defined.
Workflows are started automatically if resource type, resource format and acquisition type are defined.
Workflows emails notifications can be enabled or disabled for workflows started during the import.

add_3


jeffnm:
Reopened on development branch as #169

PR #155, opened 2017-01-24
Email reminder content

veggiematts:
Embed a concise message in the assignment email sent to users. This functionality will allow users to receive
an assignment notification with context or details specific to that workflow step and eliminate the need for
separate email communication. 

email_reminder_content


xsong9:
This is part of the our workflow enhancement, but the code change for the email reminder content was neglected to push in. The Edit Resource Step window (shown above) is opened when users re-assign a workflow step to a group by clicking on the pencil icon beside the Group in the Routing tab. We would like to include this in the 2.0 beta release.

jeffnm:
Reopened on the development branch in #168

PR #161, opened 2017-01-26
Removes "Undefinied index.." if no prefix

librariman:
If there's no http / https prefix, the isset() keeps a warning from being thrown.


jeffnm:
Reopened on development branch with #167.

PR #163, opened 2017-02-10
Support for JIT dependencies, Handle permissions errors, minor fixes

jcuenod:
Permissions errors occurred when trying to access installation scripts in folders with incorrect permissions (we now detect this and alert the user to the problem)
Minor fixes include:
 - Fixed a typo
 - Added ajax spinner (which is pretty snazzy IMHO)
 - Updated common/configuration_sample.ini
 - Updated dependencies for orgs module


jeffnm:
@jcuenod this all sounds good, but I'm not sure I fully understand why the JIT fix is needed now. We've been able to install and update many times in testing up to this point without the behavior you described in your comment that starts on line 226 in `install/installer.php`.

jeffnm:
I've been able to reproduce the JIT issue, and confirmed that this fixes it. (I never had my file permissions right before I started, who does that? ;) )

The other fixes are good too. I'll merge.

PR #167, opened 2017-02-22
Removes "Undefinied index.." if no prefix

jeffnm:
If there's no http / https prefix, the isset() keeps a warning from being thrown.

Previously #161 on master branch, reopening on new development branch.


veggiematts:
I'm ok with this.

PR #169, opened 2017-02-22
Rf47321 add fields to import

jeffnm:
See original pull request #154


jsavell:
Code merged by way of #268

PR #153, opened 2017-01-18
Import tool: uploads history

veggiematts:
history
The uploads history page. Previously uploaded files are downloadable by clicking on their filename.

history_2
Details of a previous import with the date, filename and number of imported ressources. Each imported resource is linked and their ISSN is displayed.

PR #170, opened 2017-02-22
Import tool: uploads history

jeffnm:
See original pull request #153


jsavell:
Merged by way of #269

PR #173, opened 2017-02-23
Update index.php - Remove "Undefined" notice if data is missing

librariman:
Wrapped an Html echo in a isset() IF

### DIFF
--- a/resources/index.php
+++ b/resources/index.php
@@ -64,7 +64,7 @@ include 'templates/header.php';
 	<tr>
 	<td class='searchRow'><label for='searchName'><b><?php echo _("Name (contains)");?></b></label>
 	<br />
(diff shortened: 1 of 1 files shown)


jeffnm:
Thanks for your contributions @librariman. 

We are trying a new development process out where we merge all development code into our development branch and only merge release code into the master branch. I can reopen these pull requests for you on the development branch, if you'd like, or if you are willing, you could do that. Let me know which you prefer.  

As soon as the requests are reopened on the development branch, I'm happy to merge them in light of @remocrevo review.

librariman:
Sorry for the delayed response, I’ve had so much to do lately.

I can give it a shot, I’m still not terribly familiar with GitHub so it might be good practice.

jeffnm:
That's ok, @librariman, we're all busy. I actually already switched branches and merged two of your requests, so no need to do that.

I'm still trying to understand why the remaining one isn't being flagged as a conflict since it seems to change the same line of code again.  Can you comment on #177 about how it's different than #176 was?

PR #177, opened 2017-02-24
If resourceISBNOrISSN is missing, don't display notice

librariman:
If there is nothing in the resourceISBNOrISSN variable, don't display a "Notice: Undefined..." warning.


jeffnm:
@remocrevo can you help me understand why this isn't registering as a conflict? It seems to be the exact same change as #176 which is already merged.

What am I missing?

librariman:
Unless the submitted comment counts, I have no idea.  If I find anything, I'll post it.

remocrevo:
@jeffnm Not sure what's going on here. I confirmed that this change is already in the development branch. I'm just going to close this pull request.

Issue #175, opened 2017-02-24
CORAL fails after installation with separate DB host


jamesrf:
I am running the installer with my DB on a separate host (let's call it "dbhost").  I entered this value where it said "Host".  I can get through the installer successfully and get the main menu screen.

When I select a module I get this:
'Access denied for user 'coral_user'@webserver' (using password: YES) in '.... etc

When I peek at the MySQL users table, I see that it has created a user coral_user@dbhost and granted permissions to that.  

I could not find a way to separately specify the db host from the web host.  It might be good to offer that option.  Or, the installer could use '%' as a host, although that may not be ideal as far as principle of least priviledge, so maybe have a note/advice/guide about securing your installation.

jeffnm:
Thanks for reporting this. 

To be clear, 'webserver' is the server that is serving CORAL and 'dbhost' is the server that is running mysql?

jcuenod:
Hi @jamesrf thanks for the feedback!

I just want to clarify your question - you have access denied for `'coral_user'@webserver'` but then you later say "web host" - am I correct in understanding that these are the same thing?

So your Coral install is on `http://coralserver/` but your db is on `mysql://dbserver` (not the more regular `mysql://coralserver`) and the User table should have a grant for `coraluser@dbserver` but instead it has `coraluser@coralserver`?

jcuenod:
Maybe just to follow up, the current grant is happening here: https://github.com/Coral-erm/Coral/blob/master/install/providers/have_default_db_user.php#L119

The line is:

    $db->processQuery("GRANT SELECT, INSERT, UPDATE, DELETE ON {$db_details["dbname"]}.* TO {$db_details["username"]}@{$db_details["host"]} IDENTIFIED BY '$slash_pass'");

So my understanding is that:
Instead of `{$db_details["username"]}@{$db_details["host"]}`
It should be `{$db_details["username"]}@{$_SERVER["SERVER_NAME"]}`

jamesrf:
Apologies for the delay, yes to all the above!

MySQL users are in a format user@host where host is the source of the connection (ie: coral php process running on the server that is running Apache). This allows the MySQL server to identify that the connection is coming from an authorized source

The db connection in php will also have a db host value which is so php knows what host the database is on and can connect to that host.

Hope that clarifies

jcuenod:
[...]

PR #186, opened 2017-03-04
Removing old installer code

jcuenod:


jeffnm:
@jcuenod can you provide a brief change long for this pull request? @t4k will be reviewing it, but it would be helpful to have an overview of what's new. Thanks!

jcuenod:
Hey @t4k, let me give you the highlights for each commit:

- ab0fea8: this commit literally just removes old installer code (we wouldn't want someone reading an old set of instructions or something and running it, it's also completely unused now). So this commit involves the following line changes: +0 −9,948.
- ab30990: ignore it, this is just a sync with master
- effe5d0: Handle db host being on other server (cf. #175)
- 6c8bd34: Security fix, protecting common config file on apache hosts (adding htaccess in `/common`)
- 07217c5: Improve language of error message (irrelevant, to be honest) [modified a string]
- 7ea2b1f: Updated user instructions when installing (the message was old and needed to be updated because of the way the new installer/upgrader works) [modified a string]
- d6f4752: It's worth reading #188 for this one. It began as just updating a string in a conf file but the string was non-standard and I ended up finding that `reportsModule` was preferable to `reportingModule` (which was the original) but the more deep rooted issue was the string `usageDatabase` which should be the more standard `usageDatabaseName`.

**The ones that actually need attention and should definitely be tested** are effe5d0 and d6f4752. Technically I should include 6c8bd34 but its pretty simple.

t4k:
Left a comment in #175 because I don’t think this is fixed yet.

t4k:
#175 is fixed and so I will say this whole PR is good. (@jeffnm tested the #188 parts)

jeffnm:
Thanks @t4k 

I'll merge it.

jcuenod:
Thanks, sorry for what turned out to be a gigantic pull request.

Issue #180, opened 2017-02-26
Display child records in alpha order


techsvcslib:
If you use parent/child relationships within CORAL Resources, the existing display order of child records within the parent record is not alphabetical. In fact, the display order logic is hard to determine, but alpha order is much better. See attached screenshot for illustration. This is not new to 2.0 beta--it's a longstanding "quirk"--but I think it should be fixed and shouldn't be too difficult to address.
![2017-02-26_12-31-18](https://cloud.githubusercontent.com/assets/3106487/23342429/c3131c04-fc1f-11e6-9d8f-5dc552efb4a4.gif)

PR #192, opened 2017-03-27
Display child records in alpha order

veggiematts:
Fix for issue #180


t4k:
I’ve reviewed and applied this fix locally and it does what it claims without adding any notices, warnings, or other errors.

Issue #190, opened 2017-03-16
Error on routing screen - ResourceStep missing dateArchive


jamesrf:
In Resources, I click on a resource and click "Routing", I get this error:

Fatal error: Uncaught exception 'Exception' with message 'There was a problem with the database: Unknown column 'archivingDate' in 'order clause'' in /var/www/html/resources/admin/classes/common/DBService.php:39 Stack trace: #0 /var/www/html/resources/admin/classes/common/DBService.php(73): DBService->checkForError() #1 /var/www/html/resources/admin/classes/domain/Resource.php(1954): DBService->processQuery('SELECT * FROM R...', 'assoc') #2 /var/www/html/resources/ajax_htmldata/getRoutingDetails.php(9): Resource->getResourceSteps() #3 /var/www/html/resources/ajax_htmldata.php(24): include('/var/www/html/r...') #4 {main} thrown in /var/www/html/resources/admin/classes/common/DBService.php on line 39

Looking here, this is the query being run:
https://github.com/Coral-erm/Coral/blob/b200fae47/resources/admin/classes/domain/Resource.php#L1952

However, it doesn't appear that resourceStep has an archivingDate:
https://github.com/Coral-erm/Coral/blob/b200fae47/resources/install/protected/install.sql#L471

I am using master (at 0fb5c34ecdb)

veggiematts:
archivingDate was introduced by #117 , which has been merged.
The database update is here: https://github.com/Coral-erm/Coral/pull/117/files#diff-e1bbe4cec073

jeffnm:
@veggiematts  For some reason that change to ResourceStep did not make it into the 2.0 update or install sql files.  That needs to be fixed before the 2.0.0 release.

jeffnm:
#193 fixes this for the upgrade but is missing some of the starter data for fresh installs.

PR #193, opened 2017-03-28
Error on routing screen - ResourceStep missing dateArchive

veggiematts:
Fix for issue #190

### DIFF
--- a/resources/install/protected/install.sql
+++ b/resources/install/protected/install.sql
@@ -475,6 +475,7 @@ CREATE TABLE  `ResourceStep` (
   `stepID` int(11) default NULL,
   `stepStartDate` date default NULL,
   `stepEndDate` date default NULL,
+  `archivingDate` DATETIME NULL,
   `endLoginID` varchar(200) default NULL,
   `priorStepID` int(11) default NULL,
   `stepName` varchar(200) default NULL,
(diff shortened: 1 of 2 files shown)


jeffnm:
@veggiematts thanks again for making this PR. Can you make the requested change?

resources/install/protected/install.sql
```diff
@@ -475,6 +475,7 @@ CREATE TABLE  `ResourceStep` (
   `stepID` int(11) default NULL,
   `stepStartDate` date default NULL,
   `stepEndDate` date default NULL,
+  `archivingDate` DATETIME NULL,
```
veggiematts: It is already in the install.sql file. Am I missing something here?

veggiematts:
It is already in the install.sql file. Am I missing something here ?

jeffnm:
Maybe I don't understand. The Insert statements in the update file are adding rows to several tables, but I don't see that in the install file. The starting "Any" values for those tables should be generated at install as well, or they should not be added during the update. Otherwise some will have them and others not. Right? 

I'll add line comments when I get to a computer to point out the ones I mean. The GitHub app doesn't seem to have that ability.

veggiematts:
Oh, I see. They are here as well. See upcoming review lines.

jeffnm:
Thanks! That clears up my confusion. I'll merge.

PR #195, opened 2017-03-29
Fixed #179

### LINKED ISSUE #179: Lower case letters in lefthand nav of resource record
Noticed that the default (English) word for two of the lefthand tabs in resource records does not start with a capital letter: "accounts" instead of "Accounts"; and "issues" instead of "Issues". See attached screenshot for illustration.

connordelacruz:
Fixed #179. a.showIssues and a.showAccounts were missing the relevant CSS to capitalize the first letter found in the other .show* classes

### DIFF
--- a/resources/css/style.css
+++ b/resources/css/style.css
@@ -2192,6 +2192,10 @@ a.showAccounts:before {
 		margin-right: 5px;
 }
 
+a.showAccounts {
+	text-transform: capitalize;
+}
+
 /*--- IE 11 Fix for Sidemenu Link Color ---*/
 
 .sidemenu span.link a {
@@ -2337,6 +2341,10 @@ a.showIssues:before {
     content: "\f071";
     font-size: 11px;
     margin-right: 5px;
+}
+
+a.showIssues {
+	text-transform: capitalize;
 }
 
 #issueTable img {


jcuenod:
Shouldn't this css be something more general (note especially the [content attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) in the `before` selector):

```css
a.showItem:before {
	font-family: FontAwesome;
	content: attr(data-icon);
	font-size: 11px;
	margin-right: 5px;
}

a.showItem {
	text-decoration: none;
	text-transform: capitalize;
	font-size: 12px;
}

a.showItem:hover {
	color: #ccc;
}
```

Then with the `showAccounts` link, instead of:

```html
<a class="showAccounts" data-icon="\f0c0">
```

Add the icon and use a class that can be reused:

```html
<a class="showItem" data-icon="\f0c0">
```

jeffnm:
So the idea is to move the unique data piece out of the CSS and into the data attribute to streamline the CSS. 

The question then, is whether there is javascript also using those distinct classes. If not, then yes, @jcuenod's suggestion seems to be much cleaner. 

@connordelacruz do you have time to look into it this week? If not, I'll accept this PR as an acceptable fix and we'll make an issue to look into this refactoring question later.

jeffnm:
OK, I'm merging this since it's fixing the reported issue with the same method that the code base is currently using and I don't want to hold this up for a refactor, no matter how simple it might be.

PR #197, opened 2017-03-31
Iss178 workaround

### LINKED ISSUE #178: Resource Module -> Helpful Links position
The Helpful Links area within a resource record does not position properly in the UI, overlapping important content. See attached screenshot for illustration.

connordelacruz:
Workaround that resolves #178. Set the max-width property of the URL <td> and set the style of the table to break word on word-wrap. The URL overflows outside of the <td> element if word-wrap: break-word isn't set.

A more ideal solution would likely require significant changes to styling. With the table layout, heavy in-line styling usage, explicit widths (especially when child elements had larger widths than their parent), and CSS styles marked !important, it was difficult to find the root cause and provide an alternate solution to the issue.

We tried collapsing the helpful links box and displaying it when the user hovers over it, but it still covered content and buttons when the content made the table exceed a certain width (despite having a width explicitly set in in-line styling of the table element).  We also tried moving the right panel over below the left side panel, but this too proved difficult, especially because the side menu is created several times in the same document rather than being a single menu that uses JavaScript to load content into the main content div.

We tried the negative right margin workaround that @jeffnm came up with, but it turns out -115px isn't enough when the URL width exceeds a certain point.

The changes in this pull request should deal with the issue in the cases we tested, but it seems that a more elegant solution would need an overhaul to CSS.

### DIFF
--- a/resources/ajax_htmldata/getProductDetails.php
+++ b/resources/ajax_htmldata/getProductDetails.php
@@ -57,7 +57,7 @@
 
 
 		?>
-		<table class='linedFormTable' style='width:460px;'>
+		<table class='linedFormTable' style='word-wrap: break-word; width:460px;'>
 			<tr>
 				<th width="115"></th>
 				<th></th>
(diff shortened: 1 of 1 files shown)


jeffnm:
@bdroubay were you the person who did the new CSS? Or was that someone else with your team? If so, could you have them comment on this workaround?

jeffnm:
Well, I'd really like to get rid of all the inline styling, but this is not the time or place to do that. This workaround seems to do the job, and no one has identified any issues with it, so I'm going to go ahead and merge it.

Issue #199, opened 2017-04-19
Update Footer for 2.0.0 Release Candidate


jeffnm:
The footer needs to be updated for the 2.0.0 Release Candidate release.

Issue #49, opened 2016-06-16
en_GB in LangCodes


jcuenod:
I don't have a live install right now - I'm just building an installer. I've just merged in the visual changes by Bradley Droubay and I'm getting

```
Notice: Undefined index: en_GB in /srv/http/Coral/LangCodes.php on line 15
```

Looks like a simple fix - seems it's just missing from an array.

bdroubay:
Sorry all. Because we are translating into both simplified and traditional Chinese, it was necessary to bump the length of the language codes from 2 to 5. If you add a line to the LangCodes.php file that maps en_GB to en_US, and another for the language name, it will resolve the issue. Similar action for French speaking users outside of France. We hope others will begin adding translations as well.

Let me know if you have any further questions. I'm working on merging the other modules.

Thanks,

Brad

jeffnm:
This seems like it is a bug that should be fixed in 2.0.0 if possible.

PR #203, opened 2017-04-25
Release Candidate Footer

jeffnm:
Added en_GB per #49 and updated the footer for the release candidate per #199


remocrevo:
This change looks good to me.

Issue #122, opened 2016-09-16
remove trailing whitespaces


tuxayo:
Part of https://github.com/Coral-erm/Coral/issues/77 section Coding style

Here is an idea of how to proceed.
https://stackoverflow.com/questions/149057/how-to-remove-trailing-whitespace-of-all-files-recursively

t4k:
See pull request #207

jcuenod:
Can this issue be closed?

t4k:
Yes.

PR #207, opened 2017-05-05
Issue #122: Remove trailing whitespace and some system files 

t4k:
Also correct line endings.


t4k:
I know this is huge, but we only have to do it once.

jeffnm:
Thanks for doing this!

Issue #208, opened 2017-05-05
Use .editorconfig to standardize code style


t4k:
Relates to #55, #121, & #122.

See http://editorconfig.org/ for information.

We can make various code-style rules for each type of file and set defaults. Many good editors support .editorconfig files by default or with plugins.

t4k:
This was committed in 226f5f07fb1b8d720e1a1b81d21bb16b0cd7d636.

Further discussion on our standardized coding style can be in future issues as needed.

Issue #121, opened 2016-09-16
Get rid of all DOS line endings


tuxayo:
There is a significant share of the files that have MS-DOS format line endings.

Here are some resources:
https://stackoverflow.com/questions/7068179/convert-line-endlings-for-whole-directory-tree-git/7068241#7068241
https://stackoverflow.com/questions/10629231/convert-all-eol-dos-unix-of-all-files-in-a-directory-and-sub-directories-recu

t4k:
The line endings are all set to LF in PR #207

jeffnm:
I merged #207, but there were some conflicts with some other recently merged pull requests, so it might be good to double check that all line endings are actually LF now.

t4k:
I’ll do this again to make sure we are good.

Issue #55, opened 2016-06-28
Coding guidelines: chose an indentation style


tuxayo:
There are currently [some inconsistencies](https://i.imgur.com/snm8nkn.png) leading to indentation issues between GitHub's code view and the different text editors out there.

We should choose one style and add it to the guidelines so we will gradually unify the indentation style as when new code will get written, it will fix the existing inconsistencies.

After few minutes jumping through the 70 000 lines of PHP code of Coral, I've found that > 90% of the indentation are tabs.

So tabs would be the simplest choice to avoid a mass conversion. In the case where tabs are chosen, the length should be chosen to preserve alignment (for example in a function with many params on multiple lines)

tuxayo:
Just an [example](https://github.com/ndlibersa/resources/blob/26ae17cbbc7a9b0d65786d7bd40dc6f61605c015/admin/classes/domain/Resource.php#L1189) of an alignment that rely on tab length (looks good with tabs of length 2) and [another one](https://github.com/ndlibersa/resources/blob/26ae17cbbc7a9b0d65786d7bd40dc6f61605c015/admin/classes/domain/Resource.php#L1257) that seems to be made for tabs of length 4.

t4k:
I’m not sure we need to define the length of tabs; individuals can do that how they like in their editors. If we do, we should add it to .editorconfig as proposed in #208.

If there is consensus around tabs, I’ll make the bulk change in a PR.

I feel like there needs to be a reference to the Silicon Valley episode in here somewhere, but I am failing to come up with something clever.

PaulPoulain:
t4k: that's a troll, and you jumped in ;-)
we have to chose either use TAB or SPACE for indentation. The mix is a mistake.
There are a lot of arguments against tab. One of them being that there can be a variable number of them depending on how you display them.
That's what tuxayo wrote last july.
With tab=2 spaces, you can be putting more tabs for the same indentation.
For example:
if ($bla="blabla") {
>do_this
    do_that; #with 4 white spaces
}
If you have tab=4 spaces, you'll see things properly aligned. if you have tabs=2 spaces, you'll think there's a missing tab. So you put 2, and ppl using tab= 4 spaces will think that's wrong.
So ==> use tab or spaces, but don't mix.

Note that I personally prefer, from far, spaces. But don't jump into the troll. If 90% of the code has tabs, let's use 100% tab ;-)

remocrevo:
[...]

PR #209, opened 2017-05-05
Issue #208: add an initial .editorconfig file with some defaults.

t4k:
Also relates to #55, #121, & #122.


remocrevo:
Looks good. Thanks for doing this, @t4k .

Issue #211, opened 2017-05-09
Marking a workflow complete does not save a copy in the workflow archive


scottvieira:
In testing the workflow archive I noticed that when marking a workflow complete, instead of marking individual steps in the workflow as complete, that the workflow did not show up listed in the archive.

jsavell:
We're currently experiencing this issue in master, and the #213 fix doesn't seem to be present there: https://github.com/coral-erm/coral/blob/master/resources/admin/classes/domain/ResourceAcquisition.php#L116

It looks like the fix may have gotten dropped somehow when the Workflow functionality was shifted from Resource to ResourceAcquisition: https://github.com/coral-erm/coral/commit/801c751a3db3a951a3a08e596add77300561bd21#diff-ff7cebafbff0c74330b18b40bb00d094R116

I can open a new PR restoring fix if someone can confirm that it's needed and that I'm not misreading this.

veggiematts:
As of today, I can see the fix in master.

jsavell:
Are you sure? I don't see it:
https://github.com/coral-erm/coral/blob/master/resources/admin/classes/domain/ResourceAcquisition.php#L116

I'm expecting to see lines 2004 to 2008 from this PR:
https://github.com/coral-erm/coral/pull/213/files#diff-8a0d6c3284dcd145d52d956effda2e8dL2003

I don't see isCurrentWorkflowComplete() checking for Resource completion anymore in master. Is the check for Resource completion no longer necessary or moved somewhere else?

veggiematts:
You're right, my bad. The commit containing the fix is in master, but the fix itself seems to have been lost when the code was moved from Resource to ResourceAcquisition. This is a bug.

jsavell:
Ok. I'll open a PR restoring the fix today.

PR #213, opened 2017-05-10
Issue 211: Archive workflow on restart when the workflow has been marked complete

veggiematts:
Fix issue #211


jeffnm:
This looks good.

PR #215, opened 2017-05-10
2.0.0 rc prep

jeffnm:
This PR updates the SQL to address #205 and also restores a whole bunch of images that got corrupted by accident with the trailing whitespace fix. Some of the images looked ok, but others were way off, so I just restored all of the images to be sure.

PR #217, opened 2017-05-11
Add more to beta-to-rc sql update file.

t4k:
Issue #215; additions based on git diff of install.sql from 2.0-beta & development branches.


t4k:
I think I missed this step earlier. It might be too late, but this is what might address Mang’s note.

jeffnm:
Oops, that's my fault. I thought https://github.com/caltechlibrary/Coral/commit/08516e62b00578363a28f1476634ac9a458dfc91 was included in #215 but it just referenced it.

Issue #77, opened 2016-07-08
Long term efforts to improve the codebase


tuxayo:
It's more a discussion about what should be the priorities to keep the codebase maintainable and how can we improve it incrementally.
#### Duplication between modules

Whether it's in the database or in the code itself, having now everything in the same repo allows to share stuff.

For example it seems that generic classes like DBService or other abstractions are 99% the same between modules. So for example when finishing the conversion from mysql to mysqli in Management, it would have been useful to study the possibility of merging the DB related classes with another module that was already converted. Instead of doing the same work and contributing to make them diverge in meaningless ways.
#### Coding style
- tabs in most places, spaces in some and even mixes of them (see #55)
  - alignment are often broken (in different ways depending of the editor settings!)
  - difficulties to match related blocks (if/else, especially when there is nesting) which leads to comprehension errors
- trailing whitespaces
- windows end of lines in some files
- inconsistent spacing
  - `while/if (condition) {` vs `while/if (condition){`
  - `} else {` vs `}else{`

For reasons difficult to explain, among a non negligible share of developers, these can create a lot of frustration and distract from getting work done.

One strategy that I'm experimenting is before making non trivial changes to a file (and having to deal with these style issues) I do a quick pass to fix the most blatant issues once for all in a separate commit.

For now it seems to work well.
#### Eliminating notices and warnings

When using Coral each click usually generates dozens of them and spams the logs with undefined variables, indexes, offsets, etc. Making difficult to spot one's owns mistakes, hiding existing bugs, making investigations more difficult.
[Example](https://listserv.nd.edu/cgi-bin/wa?A2=ind1607&L=CORAL-ERM&F=&S=&P=13356) of the noise added when investigating an error and how it can worry the users (ops in the libraries) [which end up disabling the error logging.](https://listserv.nd.edu/cgi-bin/wa?A2=ind1607&L=CORAL-ERM&D=0&P=14125)

It seems that there are not too many different causes and therefore gradually adding simple checks would allow to shut these messages.

Anything that could be done better? Because these strategies are a bit of improvisation given the low experience with Coral and PHP that I have.

**What are the other things that we could do to gradually improve the code and make Coral more fun to develop?**

jsavell:
### Adopt a true MVC pattern. 
- Right now, our data classes are sprinkled with view and business logic and our controllers and views are merged into one entity. This greatly complicates modifying code or adding new features.

### We should work toward using interfaces to define contracts for interacting with our data models. 
[...]

PR #219, opened 2017-05-17
Issue #77: add some checks to avoid php notices.

t4k:
Found these when working on the resources file importer.

### DIFF
--- a/resources/import.php
+++ b/resources/import.php
@@ -442,7 +442,10 @@
 						//$resource->providerText     = $data[$_POST['providerText']];
 						$resource->statusID         = 1;
 						$resource->save();
-						isset($isbnIssn_values) ? $resource->setIsbnOrIssn($isbnIssn_values) : NULL;
+						if (isset($isbnIssn_values))
+						{
+							$resource->setIsbnOrIssn($isbnIssn_values)
+						}
 						$inserted++;
 
 						// If Alias is mapped, check to see if it exists
--- a/resources/import.php
+++ b/resources/import.php
@@ -19,7 +19,7 @@
 <?php
 	// CSV configuration
 	$required_columns = array('titleText' => 0, 'resourceURL' => 0, 'resourceAltURL' => 0, 'parentResource' => 0, 'organization' => 0, 'role' => 0);
-	if ($_POST['submit'])
+	if (isset($_POST['submit']))
 	{
 		//get necessary configuration instances
 		$importConfigInstanceArray = array();
@@ -62,7 +62,7 @@
 		{
 			$error = _("Unable to upload the file");
 		}
-		if ($error)
+		if (isset($error))
 		{
 			print "<p>"._("Error: ").$error.".</p>";
 		}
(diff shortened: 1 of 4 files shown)


jcraitz:
I hope this get's merged. I was just getting ready to submit a few of these changes too.

resources/import.php
```diff
@@ -417,22 +424,25 @@ function searchForShortName($shortName, $array)
...
+						$resource->resourceFormatID		= isset($resourceFormatID) ? $resourceFormatID : '';
+						$resource->acquisitionTypeID	= isset($acquisitionTypeID) ? $acquisitionTypeID : '';
+						$resource->authenticationTypeID	= isset($authenticationTypeID) ? $authenticationTypeID : '';
+						$resource->accessMethodID		= isset($accessMethodID) ? $accessMethodID : '';
+						$resource->coverageText			= isset($data[$resourceCoverageColumn]) ? trim($data[$resourceCoverageColumn]) : '';
 						//$resource->providerText     = $data[$_POST['providerText']];
 						$resource->statusID         = 1;
 						$resource->save();
-						$resource->setIsbnOrIssn($isbnIssn_values);
+						isset($isbnIssn_values) ? $resource->setIsbnOrIssn($isbnIssn_values) : NULL;
```
remocrevo: This line is a bit strange. It would be more clear as a normal if block.

resources/import.php
```diff
@@ -305,28 +309,31 @@ function searchForShortName($shortName, $array)
...
-								$isbnIssn_values[] = $data[$columnCriterion['column']];
+								if($data[$columnCriterion['column']] != '')
+								{
+									$isbnIssn_values[] = $data[$columnCriterion['column']];
+								}
 							}
 						}
 					}
-					$deduping_count = count($resourceObj->getResourceByIsbnOrISSN($deduping_values));
+					$deduping_count = isset($deduping_values) ? count($resourceObj->getResourceByIsbnOrISSN($deduping_values)) : 0;
```
remocrevo: We really need the isset() check inside getResourceByIsbnOrISSN(), so that the input gets checked every time the function is called.

remocrevo:
Thanks @t4k for submitting this. Mostly looks great. I added a few minor suggestions. Let me know what you think.

resources/import.php
t4k: This isn’t about that function, this is about the `$deduping_values` variable.

> PHP Notice:  Undefined variable: deduping_values in /var/www/html/coral/resources/import.php on line 336, referer: http://localhost/resources/import.php

The variable is not necessarily set in any of the code above this line, so I added the `isset()` to simply check if it had been and then count the resource objects or set it to 0.

t4k: I guess some alternatives would be adding an `else` for when there is no `$dedupeCriterion['column']` or maybe even initializing an empty `$deduping_values` array before it is needed.

remocrevo: I see what you mean now. I'm happy for this to move forward as is.

Issue #223, opened 2017-05-31
submitting resources using the new api/api-client functionality


jcraitz:
Is anyone else using these new features in the release candidate?  

I fixed one issue, we were getting errors because of a path error to the email template.  I was able to do a 2 line fix for that in the Utility class (resources/admin/classes/common/Utility.php) on line 62.  Specifically, the DOCUMENT_ROOT the SCRIPT_NAME we were pulling from the server were getting munged together with an extra slash.  if SCRIPT_NAME begins with a slash, the first iteration of the for loop is just adding a slash to the document root that already has a slash on the end.

Unfortunately we're still seeing two problems.

1. When we submit a resource, it does submit in the background, but it appears to attempt a second submit and then gives us an error saying that it can't create a resource that's already there.
2. Sometimes when it doesn't give this error, it will say "The resource could not be submitted. (error: Undefined index: mailReminderDelay)".  I don't believe the form is set to create that.  Is that something that one of the Coral core functions should be handling?

Anyone else playing with this functionality?

current setup:
red hat linux 7 w/ php5.6 and apache v2.4.
using the newest version available of a mariadb cluster.

jcraitz:
Correction to the above issue.  The first problem has been fixed on our end, but the second problem is due to some needed existence checks in the Database object.  I've determined that this problem can be fixed by pull request #219.

jcraitz:
I've got a better fix coming for this issue.  Unfortunately, I still think we need a an overhaul of these 2 functions for getting the Coral Path and Coral URL.  I think it would be better to set these in the configuration.ini.  We could even suggest defaults to users when they install/upgrade using the web installer.

PR #225, opened 2017-06-01
fixing extra slash in getCoralPath

jcraitz:
Currently when $_SERVER["DOCUMENT_ROOT"] ends in a "/" and $_SERVER["SCRIPT_NAME"] starts with a slash, this function will return an invalid path.  I just added a check before it adds more slashes to the pagePath.


jcraitz:
This is the change I mentioned in the first part of issue #223.

Issue #235, opened 2017-06-13
PHP Fatal Error for Organizations on MySQL 5.7


t4k:
> PHP message: PHP Fatal error:  Uncaught Exception: There was a problem with the database: Expression #5 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'coral_organizations.OHP.parentOrganizationID’ which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by in /var/www/coral/organizations/admin/classes/common/DBService.php:39
> Stack trace:
> #0 /var/www/coral/organizations/admin/classes/common/DBService.php(71): DBService->checkForError()
> #1 /var/www/coral/organizations/admin/classes/domain/Organization.php(550): DBService->processQuery('SELECT O.organi…', 'assoc’)
> #2 /var/www/coral/organizations/ajax_htmldata.php(895): Organization->search(Array, ‘TRIM(LEADING ‘T…’, ‘’)
> #3 {main}
>   thrown in /var/www/coral/organizations/admin/classes/common/DBService.php on line 39

Diagnosed the problem and have an idea for a solution from: https://stackoverflow.com/a/34115425

Pull request coming shortly.

t4k:
This issue is actually resolved.

PR #236, opened 2017-06-13
Add item to GROUP BY for new MySQL 5.7 default compatibility.

t4k:
PHP message: PHP Fatal error:  Uncaught Exception: There was a problem with the database: Expression #5 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'coral_organizations.OHP.parentOrganizationID’ which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by in /var/www/coral/organizations/admin/classes/common/DBService.php:39
Stack trace:
#0 /var/www/coral/organizations/admin/classes/common/DBService.php(71): DBService->checkForError()
#1 /var/www/coral/organizations/admin/classes/domain/Organization.php(550): DBService->processQuery('SELECT O.organi…', 'assoc’)
#2 /var/www/coral/organizations/ajax_htmldata.php(895): Organization->search(Array, ‘TRIM(LEADING ‘T…’, ‘’)
#3 {main}
  thrown in /var/www/coral/organizations/admin/classes/common/DBService.php on line 39

Issue #235

### DIFF
--- a/organizations/admin/classes/domain/Organization.php
+++ b/organizations/admin/classes/domain/Organization.php
@@ -639,7 +639,7 @@ class Organization extends DatabaseObject {
 		$alphArray = array();
 		$result = mysqli_query($this->db->getDatabase(), "SELECT DISTINCT UPPER(SUBSTR(TRIM(LEADING 'The ' FROM name),1,1)) letter, COUNT(SUBSTR(TRIM(LEADING 'The ' FROM name),1,1)) letter_count
 								FROM Organization O
-								GROUP BY SUBSTR(TRIM(LEADING 'The ' FROM name),1,1)
+								GROUP BY SUBSTR(TRIM(LEADING 'The ' FROM name),1,1), O.name
 								ORDER BY 1;");
 
 		while ($row = mysqli_fetch_assoc($result)){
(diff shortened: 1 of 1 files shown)


organizations/admin/classes/domain/Organization.php
```diff
@@ -639,7 +639,7 @@ public function getAlphabeticalList(){
 		$alphArray = array();
 		$result = mysqli_query($this->db->getDatabase(), "SELECT DISTINCT UPPER(SUBSTR(TRIM(LEADING 'The ' FROM name),1,1)) letter, COUNT(SUBSTR(TRIM(LEADING 'The ' FROM name),1,1)) letter_count
 								FROM Organization O
-								GROUP BY SUBSTR(TRIM(LEADING 'The ' FROM name),1,1)
+								GROUP BY SUBSTR(TRIM(LEADING 'The ' FROM name),1,1), O.name
```
remocrevo: This change doesn't produce what we want. This is trying to find which letters of the alphabet are represented by existing Organizations, and how many Organizations start with that letter. This change causes "letter_count" to be 1 for each letter. I don't have an easy test environment with MySQL 5.7, but I suggest trying "GROUP BY UPPER(SUBSTR(TRIM(LEADING 'The ' FROM name),1,1))" (that is, adding the "UPPER").

t4k: Thanks for the review. It was working in my UI and I was hung up on the solution to the other issue in this pull request, but your approach is working and correct.

I’m going to remove this commit from this PR because it doesn’t have anything to do with the fatal error. I’ll add the fix to a different PR for Organizations where I am fixing some other issues.

jeffnm:
I can confirm this error occurs in mysql 5.7 and that this pr resolves that specific issue. I also don't see any immediate problems on 10.0.34-MariaDB. 

Merging.

PR #237, opened 2017-06-13
issue #445, organizations: use ANY_VALUE() around name column

t4k:
PHP message: PHP Warning:  mysqli_fetch_assoc() expects parameter 1 to be mysqli_result, boolean given in /var/www/coral/organizations/admin/classes/domain/Organization.php on line 645

Issue #77

### DIFF
--- a/organizations/admin/classes/domain/Organization.php
+++ b/organizations/admin/classes/domain/Organization.php
@@ -639,7 +639,7 @@ class Organization extends DatabaseObject {
 		$alphArray = array();
 		$result = mysqli_query($this->db->getDatabase(), "SELECT DISTINCT UPPER(SUBSTR(TRIM(LEADING 'The ' FROM name),1,1)) letter, COUNT(SUBSTR(TRIM(LEADING 'The ' FROM name),1,1)) letter_count
 								FROM Organization O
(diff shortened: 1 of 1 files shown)


t4k:
This needs work. While it fixes the warning, the query is not actually working. Investigating.

t4k:
This has become intertwined with #236.

t4k:
The underlying error from `PHP Warning:  mysqli_fetch_assoc() expects parameter
1 to be mysqli_result, boolean given in
…/organizations/admin/classes/domain/Organization.php on line 690`

is

`Expression #1 of SELECT list is not in GROUP BY clause and contains
nonaggregated column 'coral_organizations.O.name' which is not functionally
dependent on columns in GROUP BY clause; this is incompatible with
sql_mode=only_full_group_by`

Running the query directly in MySQL 5.7 will produce the error. Using the
ANY_VALUE() function suppresses the test for nondeterminism.

https://dev.mysql.com/doc/refman/5.7/en/miscellaneous-functions.html#function_any-value

t4k:
As noted in #536 the use of ANY_VALUE() is incompatible with MariaDB. This needs work.

Issue #243, opened 2017-06-21
Resource acquisition: multiple ordering in Coral


veggiematts:
In Coral, currently, when you create a resource, you also can have acquisition information. However, there can be only one acquisition information.

There are some limits related to this contraint:

- You can’t keep track of the history of subscriptions (sub start/end, order number and system number, as well as purchasing sites are unique)
- The cost history is not linked to the ordering information, it’s linked to the resource itself.

As you acquire a resource periodically, the first change to do is to enable multiple (yearly) ordering in Coral. Each order can be linked to a different vendor/organization, as a library can change it’s vendor each year.
A given resource at title level can also be available in different packages (parent resource), from different vendors. For each resource for a given package, the conditions (license) can differ. We also need to cover this case.
Last, but not least, in a consortium environment, a resource can be bought for some libraries (authorized site), not for all. And two sites can buy a given title from different sources.

PR #244, opened 2017-06-21
Resource acquisition: multiple ordering in Coral

veggiematts:
This patch allows multiple ordering in Coral, see Issue #243 for why it is needed.

Test plan:
- Run `resources/install/protected/update_Orders.sql`, it will create the new ResourceAcquisition table and change existing tables to be linked to ResourceAcquisition, create a new default order for every resource in the database, update resource data to be linked to this default order, and delete now unused fields from the Resource table.

You now have a new 'Order' panel for every resource in your database:
orders_1

You can now create a new order for your resource. Note that when you have more than one order, a dropdown list will appear, right after the resource title, allowing you to select which order you want to display informations for:

orders_2

You can also clone an existing order, allowing you to easily create a new order without having to input all informations again.

Now all informations in the panels below Product will be order specific.

You can have order-specific and resource-inherited contacts:
orders_3

Please note that even though code and database changes are quite important, current users that do not need this feature will see almost no change to the interface when using the single default order.


veggiematts:
Fixed DOS line ending and trailing spaces for the diff to be easier to read.

t4k:
My environment:
> Server version: Apache/2.4.18 (Ubuntu)
> mysql  Ver 14.14 Distrib 5.7.18, for Linux (x86_64) using  EditLine wrapper
> PHP 7.1.6-1\~ubuntu16.04.1+deb.sury.org+1 (cli) (built: Jun  9 2017 08:26:34) ( NTS )
>    with Zend OPcache v7.1.6-1\~ubuntu16.04.1+deb.sury.org+1

I started testing this. Some notes…

1. I got an error running the SQL in `update_Orders.sql`:
> ALTER TABLE `ResourceAcquisition` ADD INDEX(`licenseID`)
> Error in query (1072): Key column ‘licenseID’ doesn’t exist in table 

2. The `update_Orders2.php` file should be run directly from within the `protected` directory on the command line. It didn’t work when I tried running it from the project root.

3. PHP Fatal Error, and other warnings and notices:
> PHP Fatal error: Uncaught Exception: There was a problem with the database: Expression #​3 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'coraltest_resources.AT.shortName’ which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by in …/resources/admin/classes/common/DBService.php:39 Stack trace: #0 …/resources/admin/classes/common/DBService.php(73): DBService->checkForError() #​1 …/resources/admin/classes/domain/Resource.php(799): DBService->processQuery('SELECT R.resour…', 'assoc’) #​2 …/resources/ajax_htmldata/getSearchResources.php(35): Resource->search(Array, ‘R.createDate DE…’, ‘0, 25’) #​3 …/resources/ajax_htmldata.php(24): include(‘…’) #​4 {main} thrown in …/resources/admin/classes/common/DBService.php on line 39

I got past the fatal error by changing line 724 in …/resources/admin/classes/domain/Resource.php to: `$groupBy = "GROUP BY R.resourceID, acquisitionType";`

> PHP Warning: Declaration of GeneralSubject::save() should be compatible with DatabaseObject::save($new = 0) in …/resources/admin/classes/domain/GeneralSubject.php on line 0

See https://github.com/Coral-erm/Coral/blob/development/resources/admin/classes/domain/GeneralSubject.php#L84 for the `save()` function that doesn’t match the edited `save($new = 0)` function.

I added `$new = 0` into the `save()` function in the `GeneralSubject.php` file to get past the warning.

The rest of these notices happened while working in the UI experimenting with adding orders.

> PHP Notice: Undefined index: resourceAcquisitionID in …/resources/resource.php on line 23

> PHP Notice: Undefined variable: tableWidth in …/resources/ajax_htmldata/getOrdersDetails.php on line 26

> PHP Notice: Undefined index: resourceAcquisitionID in …/resources/ajax_forms/getOrderForm.php on line 2

> PHP Notice: Undefined index: op in …/resources/ajax_forms/getOrderForm.php on line 4

> PHP Notice: Undefined variable: organizationName in …/resources/ajax_forms/getOrderForm.php on line 96

> PHP Notice: Undefined index: subscriptionAlertEnabledInd in …/resources/ajax_processing/submitAcquisitions.php on line 27

In the UI “There was a problem with the database: Field ‘subscriptionStartDate’ doesn’t have a default value” was also displayed when no date was entered.

After entering a date and clicking Submit I got the same PHP Notice. I then had to check the Enable Alert box for the window to go away. At that point I had saved two copies of the new order. It seems that the first one saved but the modal window didn’t go away.

When trying to Clone the modal does not ever go away, even after checking the Enable Alert box, and additionally I get the following message in the UI: 
> There was a problem with the database: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ’’ at line 1

When editing an order, I have to check Enable Alert as well, but the modal does go away after doing that.

4. It seems that once an order is created there is no way to delete it. I think this should probably be addressed.

t4k:
Also, what’s the plan for users to run the `update_Orders2.php` script on their systems. I have to imagine not everyone will be able to get at their command line so we should develop an interface for this.

t4k:
In showing these potential changes to my colleagues, @lnarizny and @caltechheather, we thought that when changing fields and data structure of the Resources table(s) we really should be sure to update the code for the Importer as well. If these changes go in, users will not be able to import multiple orders through the importer.

techsvcslib:
Not sure we should do that at this stage given this new feature is optional. This is a big change, we’ve got some needs to iron out the importer as it stands anyway, and I’d rather wait to see how this works.

Steve

On Aug 10, 2017, at 2:00 PM, Tommy Keswick <notifications@github.com<mailto:notifications@github.com>> wrote:


In showing these potential changes to my colleagues, @lnarizny<https://github.com/lnarizny> and @caltechheather<https://github.com/caltechheather>, we thought that when changing fields and data structure of the Resources table(s) we really should be sure to update the code for the Importer as well. If these changes go in, users will not be able to import multiple orders through the importer.

veggiematts:
@t4k Thanks for your comments ! 

I made a QA fix commit (I couldn't reproduce the clone error, however, and the option to delete an order has yet to be done)

I totally agree that the modifications should be part of the new installer, but I'm not familiar with it. Is there somewhere a documentation or a how-to that might help me understand how to add new modifications to it ?

veggiematts:
And about the Import tool, it currently creates a default order when importing resources.

There are a lot of things the import tool does not handle right now, and I'm not sure importing orders would be the most needed feature to add to it.

veggiematts:
The update process is now available through a single SQL file so it can be used in the installer.

veggiematts:
Test plan updated.

t4k:
@veggiematts: I'd like to begin testing this one. Could you resolve the conflict in this PR?

veggiematts:
@t4k done !

t4k:
I've tested this again and got a few of the same problems as the first time:

> PHP Notice: Undefined index: resourceAcquisitionID in …/resources/resource.php on line 23

> PHP Notice: Undefined variable: tableWidth in …/resources/ajax_htmldata/getOrdersDetails.php on line 26

> PHP Notice: Undefined index: resourceAcquisitionID in …/resources/ajax_forms/getOrderForm.php on line 2

> PHP Notice: Undefined index: op in …/resources/ajax_forms/getOrderForm.php on line 4

> PHP Notice: Undefined variable: organizationName in …/resources/ajax_forms/getOrderForm.php on line 96

> PHP Notice: Undefined index: subscriptionAlertEnabledInd in …/resources/ajax_processing/submitAcquisitions.php on line 27

After entering a date and clicking Submit I that last PHP Notice. I then had to check the Enable Alert box for the window to go away. At that point I had saved two copies of the new order. It seems that the first one saved but the modal window didn’t go away.

When trying to Clone the modal does not ever go away, even after checking the Enable Alert box, and additionally I get the following message in the UI: 
> There was a problem with the database: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ’’ at line 1

When editing an order, I have to check Enable Alert as well, but the modal does go away after doing that.

t4k:
Update to the comment above.

The modal issues arise because on my test environment I have PHP Error Reporting turned on, and this interferes with the modal window behavior because notices are printed all over the screen.

I have tested now with PHP Error Reporting turned off and the only UI issue that remains is the message when trying to clone an order record:
> There was a problem with the database: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ’’ at line 1

A cloned record is created, but the modal does not close. Clicking Cancel and then refreshing shows the cloned record in the select list. (Note, I am using MySQL 5.7 in this test environment.)

All that said, is there a way to harden the code so that it can function even if errors are printed on the screen? I really think we should do this, because it seems very fragile otherwise.

And also, the underlying PHP Notices still need to be addressed. Turning off Error Reporting is not a solution.

veggiematts:
I've added a small patch that prevents the PHP Notices.
However, I still have to investigate the clone order bug, which seems to be a very peculiar one.

t4k:
More on the clone issue:

If I clone an order record, and get that database error, and then cancel out of the modal, there is a brand new cloned *resource* on the resources list. I don't think this is the intention.

veggiematts:
The clone issue should be fixed.

t4k:
The SQL error is fixed with the latest updated.

The issue from my last comment still remains. I have looked further at it and can report more details.

I believe the issue is with the query that generates the index page list of resources (/resources/index.php). When there is a resource that has multiple orders, that resource is duplicated in the index list, one for each order record in the system. Maybe there has to be a DISTINCT put on that query or something similar?

Steps to reproduce in the sandbox… http://user01-coral-sandbox.test.biblibre.eu/sandbox.php

1. Start the sandbox with Pull Request #244.
2. Add a New Resource.
3. Go to the Orders vertical tab of that resource.
4. Create a New Order.
5. Go to Resources home.
6. See the same resource repeated twice; note that both link to the same URL.
7. Click into the resource.
8. Go to the Orders vertical tab.
9. Clone an Order.
10. See the same resource repeated three times; all linking to the same URL.

veggiematts:
fixed.

t4k:
This will not work for MySQL 5.7. We get the same issue as I noted in #235.

I see more of what the problem is now. Because the order records are where the Acquisition Type is stored and we also show Acquisition Type on the index page in the table, we are going to have the problem of what to put in that field for the Resource.

This might need to discussed among Steering Committee members if we need to remove the Acquisition Type field from the index table or if we need to allow multiple Resource records if the Acquisition Type changes between orders.

veggiematts:
Ok, so here's what I would to keep MySQL 5.7 happy and to keep consistency in user display.
This way, the group by is not ambiguous, so MySQL shouldn't complain.
And for the user: 
 - if all orders for a resource have the same acquisition type, it will be displayed as usual.
 - if some orders for a resource have different acquisition types, they will be displayed separated by slashes ('Free / Paid / Trial', for example). 

Is that ok with you?

t4k:
I'll test that nothing is broken and give my OK on the functionality.

If others see this, I would defer to those who use that screen the most, since it is kind of a UI change.

PR #246, opened 2017-06-21
Make Coral work with Shibboleth

fondrenlibrary:
Add back missing Logout buttons when not using Coral Auth module.
Introduce new configuration parameter - idpLogout in configuration_sample.ini of each module to hold Shibboleth IDP Logout URL.
Add htaccess template file by the name of htaccess_sample4shib into the top directory of  each module.

### DIFF
--- a/licensing/admin/configuration_sample.ini
+++ b/licensing/admin/configuration_sample.ini
@@ -9,6 +9,7 @@ resourcesDatabaseName=
 managementModule=Y
 useTermsToolFunctionality=N
 remoteAuthVariableName="_SERVER['REMOTE_USER']"
+#idpLogout="https://idp.youruniversity/idp/profile/Logout"
 
 [database]
 type = "mysql"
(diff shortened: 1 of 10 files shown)


jeffnm:
I wonder if it would be better to make the Auth module handle Shibboleth integration, like it does with LDAP now, rather than going back to .htaccess files. 

Also, I'm concerned about adding a logout button setting to the configuration. It seems to me that we should not have to add a flag like that to every configuration file. I don't have another suggestion at the moment though.

I appreciate you working on getting Shibboleth support! It's definitely needed!

fondrenlibrary:
In order to completely log out of Shibboleth Session,  the associated IDP session must be ended using the IDP logout URL which can be hard coded or stored in the configuration file. Shibboleth is different from LDAP and according to my current knowledge and our test, using htaccess is the easiest way to go.

fondrenlibrary:
Actually, my 2nd thought suggests using htaccess may not be necessary, but the idp Logout URL has to be used, according to my current knowledge which is not sufficient.

fondrenlibrary:
How about only using  htaccess in the auth folder ?

jeffnm:
I don't know enough about Shibboleth to be very helpful, sorry. I expect I'll be learning about it later this year, so hopefully I learn something from what you come up with.

fondrenlibrary:
In order to  use .htaccess only with the Auth module,  index file, configuration.ini and maybe some other ones need to be modified  to handle server parameter REMOTE_USER which is set by Apache after a user is authenticated via Shibboleth.

jeffnm:
We still need to add support for this, but I'm still not in a position to test it.

PR #250, opened 2017-06-30
Fix usage module reporting link

jeffnm:
The name reportingModule setting name changed in 2.0 to match the other setting name conventions, but the corresponding menu check did not get updated.

### DIFF
--- a/usage/templates/header.php
+++ b/usage/templates/header.php
@@ -196,7 +196,7 @@ $coralURL = $util->getCORALURL();
         </div>
     </a>
 
-    <?php if ($config->settings->reportsModule == "Y") {
+    <?php if ($config->settings->reportingModule == "Y") {
     ?>
     <a href='../reports/' target='_blank'>
         <div class="main-menu-link">


t4k:
Tried testing this and it didn’t quite go as expected. The key for the configuration option still gets set as `reportsModule` in `/usage/admin/configuration.ini` because of https://github.com/Coral-erm/Coral/blob/master/reports/install/reports.php#L8

t4k:
If the installer code for Reports is updated to set the `$MODULE_VARS = ['uid']` to “reporting”, what would happen to existing installations that have a Usage `configuration.ini` file with the lines
```
reportsModule = "Y"
reportsDatabaseName = "coral_reports"
```
in them? Will the installer/updater take care of updating the `configuration.ini` file?

My concern is that as soon as the `header.php` file from this PR gets added to an existing site, the menu link will not work for those that have an old `configuration.ini` file.

@jcuenod ?

jcuenod:
In theory if the installer code is updated then only new installs would be fixed.
To fix this we would need an update script to handle this particular problem (although right now the updater is incomplete and so how/whether we're going to write and update the common conf file is not certain yet)

That said, I should add that I don't recall testing using folder names that didn't match UIDs (I hope that I was diligent enough to code as though they needn't match but that strikes me as a possible problem).

I also recall discussing whether it would be okay to change the conf file for usage because there was inconsistency between usage and something else (I don't recall what) in the way it referred to other modules (it may have been just within usage's own conf file). It seemed to make more sense to use the UID from reports. So my gut response to this issue is that we _want_ "reports" to be set in the conf file rather than "reporting", we just need to figure out why Jeff's fix would not be functioning correctly for the menu check.

jeffnm:
It looks like I made a mistake. I thought that the 2.0 upgrader had changed the Usage `configuration.ini` file to use `reportingModule` instead of `reportsModule` to match the name of the database. I was apparently wrong about that, it looks like we instead changed the database name.  https://github.com/coral-erm/coral/commit/d6f47524de274609036574d5b9ad056a3b17ac0f#diff-966673a22f806dbf3a4831e86ac51bc1
So this pull request is a mistake based on our installation.

PR #269, opened 2017-08-04
Update PR #170 to resolve merge conflicts.

t4k:
There were merge conflicts in PR #170 (old PR #153) that needed to be resolved. Also included an isset() call to reduce notices in the logs.

This is the second part of the review and approval of everything in #8.

### DIFF
--- a/resources/admin/classes/common/DatabaseObject.php
+++ b/resources/admin/classes/common/DatabaseObject.php
@@ -221,7 +221,7 @@ class DatabaseObject extends DynamicObject {
 			if (isset($this->attributes[$attributeName]))
 			{
 				$value = $this->attributes[$attributeName];
-				if ($value == '' || !isset($value)) {
+				if (($value == '' || !isset($value)) && $value !== 0) {
 					$value = "NULL";
 				} else {
 					$value = $this->db->escapeString($value);
--- a/resources/admin/classes/domain/ImportHistory.php
+++ b/resources/admin/classes/domain/ImportHistory.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+**************************************************************************************************************************
+** CORAL Resources Module v. 1.0
+**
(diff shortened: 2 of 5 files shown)


t4k:
One comment about the functionality in this PR that should be added to its own issue once this is merged:

Imported CSV files uploaded to the attachments directory are silently overwritten if they have the same filename as an existing file in that directory. This could cause confusion because the links on the Import History page to the previously uploaded files with the same name(s) would all link to the most recently uploaded file.

I don’t think this is a blocking issue, but would certainly require a Known Issues note and some documentation.

scottvieira:
@t4k How hard would it be to just fixed the code so that it doesn't overwrite the existing file without user intervention?

techsvcslib:
This is my preference also.

Steve

On Aug 4, 2017, at 11:37 AM, Scott Vieira <notifications@github.com<mailto:notifications@github.com>> wrote:


@t4k<https://github.com/t4k> How hard would it be to just fixed the code so that it doesn't overwrite the existing file without user intervention?

t4k:
@scottvieira @techsvcslib I am trying to concentrate on getting this pull requests up to speed right now so we can at least merge them.

Adding file renaming or rejecting would be a sizable amount of code that we really should develop for the project as a whole. There are many places where file uploads happen in CORAL and we should write code that will handle this globally when we have a chance. An issue should be opened for it.

The thing is, this silent overwriting behavior is how it was already working. By exposing upload history, we are just making records for data that was sort of ephemeral before.

scottvieira:
@t4k @techsvcslib I agree.  Putting this down as an issue works for me.  I don't know PHP, but I'm surprised there isn't existing code or a function for this.

t4k:
@scottvieira Yes, there are existing patterns to use for something like this, but there is not just something one can drop into existing code that will work with the existing application without considering how to integrate it with the rest of the codebase.

PR #171, opened 2017-02-22
Import tool: Preview the mappings

jeffnm:
See the previous pull request #152

PR #271, opened 2017-08-07
Update PR #171 to resolve merge conflicts.

t4k:
There were merge conflicts in PR #171 (old PR #152) that needed to be resolved. Also included a number of fixes to reduce warnings and notices in the logs.

This is the third part of the review and approval of everything in #8.

### DIFF
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -59,6 +59,7 @@ class Resource extends DatabaseObject {
 		$i = 0;
 
 		if (!is_array($isbnOrISSN)) {
+			if ($isbnOrISSN === null) return;
 			$value = $isbnOrISSN;
 			$isbnOrISSN = array($value);
 		}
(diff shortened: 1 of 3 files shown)


t4k:
A comment about the functionality of this pull request:

There are a number of UI additions that could probably use some work. They aren’t devastating, so I would move forward with this regardless.

The first is on the screen where the mapping occurs. The columns table is laid out horizontally which is very unreadable with a spreadsheet that has a large number of columns. On this screen there is no good reason to be using a horizontal table to display this information. We could have a two-column table with Column Name and Column Number. This would make the reference so much easier to use when mapping.

The second issue is the Preview screen. I’m not actually sure we can do anything about the way a table will expand very wide if there are a large number of columns. Maybe we consider a table that scrolls within the page?

The third issue would be the Mapping table below the preview. When the preview table is very wide, this table expands to 100% just like it, even though the data is much more concise. I would argue like in my first issue above that this data can be displayed in a vertical table.

Also, in the Mapping table there is a column defined called “File column” that has no code that will ever populate it with anything.

t4k:
The other thing about this particular pull request is that we will have merge conflicts with the other two related to issue #8. #268 and #269 can be merged into development cleanly together, but this PR will cause conflicts with both of those. Once we decide the order of merging, more work will need to be done.

PR #196, opened 2017-03-30
Merge common js files between modules

veggiematts:
This is a rebase of issue #84 , previously pull request #88 on devel branch.

Issue #84, opened 2016-07-13
Merge js libs between modules (and maybe our own js + css)


tuxayo:
In `js/plugins` directories.

We have 8 jquery.js, and 8 Gettext.js

Which prevents excluding them when searching through all the code.

tuxayo:
When #88 will be merged, there will still be:

- [ ] identical files that are common between modules (maybe only 2 duplicates)
- [ ] files that have some differences that should be merged manually and tested
- [ ] 4 different versions of jQuery
    - 1.3.2
    - 1.4.4
    - 1.8.0
    - 1.11.3

tuxayo:
As a side note, I used the [FSlint](https://www.pixelbeat.org/fslint/) program to find duplicate files.

PR #272, opened 2017-08-08
Update PR #196 to resolve merge conflicts.

t4k:
Issue #84. There were merge conflicts in #196 which needed to be resolved. I merged with line endings ignored, which probably made the conflicts seem minimal to non-existent.


t4k:
In testing there were a few visual changes, e.g. font sizes appear larger in some places.

I think consolidating these files is far more important for ease of future development than some font sizing issues.

t4k:
Hmm. Now that I’m looking at this one vs. #196 the number of files changed is very small. I will have to look into this more.

t4k:
The number of files affected changed only because of the way Git was renaming or deleting them, I believe.

Issue #275, opened 2017-08-14
Switch to ChromeDriver instead of PhantomJS


t4k:
The maintainer of PhantomJS stepped down earlier this year: https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE

We should probably move to ChromeDriver or something else. I have been testing it out locally and will put up something soon to upgrade the automated test suite.

PR #278, opened 2017-08-14
Issue #275: Update files for latest Codeception using ChromeDriver.

t4k:


t4k:
Created new #305 with squashed commits for a cleaner history.

PR #283, opened 2017-09-01
Issue 282 french translation

### LINKED ISSUE #282: French translation half-working
The first two letters of every string in the french translation are missing in index, auth and resources modules.

veggiematts:
This PR fixes #282 .

I don't exactly know why the translations were incorrectly displayed, but re-generating the mo files with msgfmt fixed the problem.


t4k:
I was unable to reproduce the initial problem on my test machine, but I will trust this takes care of it.

PR #288, opened 2017-09-08
Use composer to manage API dependencies

### LINKED ISSUE #287: Resource: Use Composer to manage API dependencies
This will allow the API feature in the Resource module to be used without manual download and installation of the Flight dependency.

Composer is already being used to manage dependencies for the API Client example and for the Codeception test suite.

jsavell:
Resolves #287

### DIFF
--- a/resources/api/INSTALL
+++ b/resources/api/INSTALL
@@ -1,4 +1,4 @@
 The API uses Flight, a fast, simple, extensible framework for PHP.
-Just install Flight in this directory, so Flight.php is available like this: resources/api/Flight/flight/Flight.php
+Install Flight in this directory with composer, so autoload.php is available like this: resources/api/vendor/autoload.php
 
 http://flightphp.com/
--- a/resources/api/composer.json
+++ b/resources/api/composer.json
@@ -1,0 +1,5 @@
+{
+    "require": {
+        "mikecao/flight": "^1.3"
+    }
+}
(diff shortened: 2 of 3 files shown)


jcraitz:
Thank you for submitting this one.  Was going to suggest we do this.

veggiematts:
Since we're already using composer in Coral, I'm in ok with using it also for the API.
However, merging this will break the API for people already using it.
How do we handle this ?

xsong9:
We are using the Resources API on a daily basis now. Is this change going to break this one? If so, please hold off on the change until you have a solution. Thanks!

jcraitz:
At least for us it's not going to be a difficult change at all.  If people were using it as it was documented before, then they downloaded the package separately in the Flight folder.  Making the proposed change will add a separate vendor folder.  No conflicts there.  And the maintainer simply deletes the old Flight folder.

xsong9:
Good to know it will not break our API. Is someone going to test it before production? We are ok if the test goes through.

jcraitz:
I have tested the same change in a previous version.  It's fine.

jsavell:
@veggiematts It does require re-following the updated API installation instructions, but that doesn't seem like too much of a burden.

If it is too much of a burden, we could support both Composer and manual dependency approaches simultaneously by having the code check for the presence of Flight under `vendor`, fall back to check the previously recommended Flight manual install location.

Depending on the result, the code can conditionally include the autoloader or a direct include of Flight.

t4k:
I added a documentation label to the parent issue. The reconfiguration of the dependency should be added to the Release Notes.

Issue #289, opened 2017-09-12
SQL Typo in Update_2.0.0 sql file


jcraitz:
There is a typo on the first line of the file.  'ResourcePayment' should be 'ResourceType'.  This typo fails to create the 'includeStats' column in the ResourceType table.  When upgrading from v1.4, this typo will give the web upgrade script the following error:
`Unknown column 'includeStats' in 'field list'
For statement: INSERT INTO 'ResourceType' ('resourceTypeID', 'shortName', 'includeStats') VALUES (NULL, 'Any', NULL)`

I'm creating a single line pull request to fix this issue.

PR #290, opened 2017-09-12
fixing typo in update script.

jcraitz:
Fix for #289.


jcraitz:
above comment on commit should be "fixing typo in update script."

There was a typo in my message about fixing a typo. <sigh>

PR #295, opened 2017-09-15
Delete info.php

jeffnm:
We should not be including `phpinfo();` calls in the repository. It will cause unwary users to accidentally expose server configuration information.


jcuenod:
A quick grep reveals only `usage/info.php`

PR #299, opened 2017-09-20
Move shared module directory code to a shared file and support autoloading module interfaces

### LINKED ISSUE #297: Modules should share a common directory file
Currently, each module defines its own `directory.php` where global functions and configuration are handled.

Much of this code is common across all modules but is duplicated, rather than shared.

Each module directory file should include a global `common_directory.php` file where the common code can be shared, while still preserving the ability to add module specific code to the module directory file.

### LINKED ISSUE #298: Update the autoloader to support module interfaces
CORAL is on its way to employing interfaces, which is good, but we should avoid requires and includes by updating the existing autoloader to find them automatically when placed in an `interfaces` subdirectory under a module's `admin` directory.

jsavell:
Resolves #297 - CORAL should behave the same, just with improved code reuse and maintainability

Resolves #298 - Module classes should be able to implement any interfaces defined under `[module]/admin/interfaces` without any manual require/include in that class file.


jcuenod:
Nice, this is going to be a long process...
What is the reason that you didn't move `format_date()` to the common file?

jsavell:
@jcuenod 

I just didn't realize how many modules were using `format_date()`

jcuenod:
Okay then also, regarding `debug` in `common/common_directory.php`:
1. Is it being used anywhere (as far as I can tell there are a couple of places that it's commented out)
2. Is it something that should be in the code base (I don't know what the feeling about that is).

veggiematts:
Tested on my instance, works for me.

t4k:
I question the filename `common/common_directory.php`. It seems redundant and I imagine that we will eventually have many files within the `common` directory. Are we going to prefix them all with `common_`?

t4k:
Is there a reason the `LangCodes.php` file cannot be added to the `common` directory as well? I know the root `index.php` file uses it, but it would be nice if the root directory were cleaner and the common `LangCodes.php` file was actually located sensibly in `common`.

jsavell:
I don't have an opinion on the `common_directory.php` file name. I can change it do `directory.php`

I didn't investigate moving the root `LangCodes.php` file from where it was already located.

veggiematts:
This patch breaks the API.
Please merge https://github.com/biblibre/Coral/commit/143d1aaf6249b4e28b56767a07dd076b9f73c602 to your branch to fix the problem.
Otherwise, it seems all good.

jsavell:
Thanks for catching and fixing that @veggiematts . The followup PR is merged in.

jsavell:
The new commits introduced by the followup PR merge all match hashes of commits already in the development branch, so I think it's just an artifact of how Github is presenting this older PR.

veggiematts:
As far as I'm concerned, we're good to go.

PR #300, opened 2017-09-25
Utility class and api_client form fixes

jcraitz:
fixing #223.
This should fix any unexpected slashes in coral paths or urls when calling the getCORALPath or getCORALURL utility functions from the API.  One small addition is a $defaultMessage variable initialization to avoid throwing errors when the template file is opened.

This should also fix some fundamental flaws with the example api_client form.

### DIFF
--- a/organizations/admin/classes/common/Utility.php
+++ b/organizations/admin/classes/common/Utility.php
@@ -54,7 +54,7 @@ class Utility {
 
 	//returns file path up to /coral/
 	public function getCORALPath(){
-		$pagePath = $_SERVER["DOCUMENT_ROOT"];
+                $pagePath = rtrim($_SERVER['DOCUMENT_ROOT'],'/\\').'/';
 
 		$currentFile = $_SERVER["SCRIPT_NAME"];
 		$parts = Explode('/', $currentFile);
--- a/resources/admin/classes/common/Utility.php
+++ b/resources/admin/classes/common/Utility.php
@@ -54,12 +54,14 @@ class Utility {
 
 	//returns file path up to /coral/
 	public function getCORALPath(){
-		$pagePath = $_SERVER["DOCUMENT_ROOT"];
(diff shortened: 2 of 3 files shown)


veggiematts:
Could you provide a test plan for this PR?
I've tried displaying the result of getCORALURL and getCORALPath in a php test file with and without your patch, and the results are the same for me:

http://coral.local/
/home/coral/www/

jcraitz:
I added a few extra little fixes to the form to make it a bit less brittle. 

- added a couple (isset or !empty) checks to make sure a variable was available before calling it.
- added body tags
- fixed a typo

Here's a Test Plan @t4k, thanks for reviewing! 
- Make sure your account is set up to be notified when a resource is submitted (Note, you may need to add the 'approved' or 'needs approval' for acquisition types and set up an associated workflow with a notification step for your user)
- Submit a resource for that workflow using this form from the current development branch
- You will get an email with an error similar to the one attached.

screen shot 2018-01-11 at 11 41 58 am


The notification creation from the file template assumes that it is being initiated from different file/URL paths.  I still believe, in the future, we should set up these functions in the Utility Class to pull the file & URL paths from the config file, but this solves the short term problem mentioned in #223.

jcraitz:
Would it be possible to get this merged to development?  I think it would make testing a little easier as the current form in development is broken with respect to sending notification emails and some PHP problems (functions calling $headers and $footers out of scope).

veggiematts:
Starting a new workflow on resource submission is currently broken in devel, so it's kind of pointless to merge this to devel now, but I'll test this PR as soon as a PR is submitted for fixing the start of a new workflow on resource submission (see #345 )

jcraitz:
I believe this PR should be ready for testing now, @veggiematts.

jcraitz:
No need to merge this anymore. it was merged as part of #344.

Issue #306, opened 2017-10-10
Tests should check for PHP Notices and Warnings


t4k:
Because CORAL cannot work without JavaScript, we can’t use functional automated tests to check for PHP Notices and Warnings. We’re stuck with acceptance tests.

We can check for the words “Notice“ or “Warning“ printed on the page within `<b>` tags when running acceptance tests.

rtatterson:
Developer help needed.

PR #305, opened 2017-10-10
Issue #275: update to chromedriver

t4k:
This PR replaces #278; I squashed the commits for a cleaner history.

# REVIEW PROCEDURE
Follow the instructions in `TESTS.md` on a local/test system and see if they work for you.

I’ve spent a good amount of time trying this in different environments. I have it working in Ubuntu 14.04 with PHP 5.6 and 16.04 with PHP 7.1.

Please Note: The tests will likely fail, but the testing environment should be set up properly. I will file more issues for fixes to the existing tests.

PR #307, opened 2017-10-10
Issue #306: Check for PHP Notices and Warnings before and after tests.

t4k:
Addresses #306.

This should be applied in conjunction with #305 because the automated tests are broken until #305 is applied (the `configuration.ini` file path is wrong).

## REVIEW PROCEDURE
Run an automated test with a known PHP Notice or Warning and the test will fail. For example, currently the `LoginCept.php` test will fail with “Notice” printed on the page inside a `<b>` tag.


t4k:
I’m going to propose a new way to do this. Closing.

PR #310, opened 2017-10-18
Fix non-ascii apostrophe for translations

veggiematts:
Non-ascii characters in the code strings make gettext unhappy:

```
find . -type f -iname '*.php' -o -iname '*.js' | grep -v "plugins" | xgettext -j -f -
xgettext: Non-ASCII string at ./ajax_forms/getImportConfigForm.php:209
```
I've only found one occurrence of a non-ascii character so far, here's the patch to replace it with its ascii equivalent.

### DIFF
--- a/resources/ajax_forms/getImportConfigForm.php
+++ b/resources/ajax_forms/getImportConfigForm.php
@@ -206,7 +206,7 @@
 
 			<fieldset>
 			<legend><?php echo _("Organization Name Mapping");?></legend><div id='resource_organization'>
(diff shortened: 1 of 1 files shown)


t4k:
I found this: https://www.gnu.org/software/gettext/manual/html_node/po_002fRules_002d_002a.html

It seems to indicate you can configure gettext to deal with the quotes. I don’t know anything about this, but I really prefer typographical apostrophes and quotations marks if at all possible. It’s much more professional.

This tutorial https://lingohub.com/blog/2013/07/php-internationalization-with-gettext-tutorial/ has examples with typographic apostrophes in it and it does not note anything special about them.

veggiematts:
Okay, we should use `xgettext --from-code=UTF-8 -j -f` then, assuming special characters in the source code will all be UTF-8.

PR #312, opened 2017-10-19
Issue #275: Update automated testing suite to use ChromeDriver

t4k:
This PR replaces #305 and #278, and addresses #275.

Note: The version of chromedriver is locked at 2.32 right now due to facebook/php-webdriver#469 which is for implementing the latest W3C specification for WebDriver.

I’ve removed all the functional and unit test files to lessen confusion because we do not have and cannot run any of those tests.

# REVIEW PROCEDURE
Follow the instructions in `TESTS.md` on a local/test system and see if they work for you.

I’ve spent a good amount of time trying this in different environments. I have it working in Ubuntu 14.04 with PHP 5.6 and 16.04 with PHP 7.1.

Please Note: The tests will likely fail, but the testing environment should be set up properly. I will file more issues for fixes to the existing tests.

PR #313, opened 2017-10-19
Issue #306: Add a custom action to check for PHP messages.

t4k:
Addresses #306. (Replaces PR #307.) 

This should be applied in conjunction with #312 because the automated tests are broken until #312 is applied (the `configuration.ini` file path is wrong).

This PR adds an action that can be included in tests called `checkForPhpMessages()` which ideally should be used after `amOnPage()` and `click()` assertions, like:
```php
$I->amOnPage("/auth/");
$I->checkForPhpMessages();
…
$I->click("#loginbutton");
$I->checkForPhpMessages();
```

## REVIEW PROCEDURE
Add the `$I->checkForPhpMessages();` assertion to an automated test with a known PHP Notice or Warning and the test will fail. For example, currently the `LoginCept.php` test will fail with “Notice” printed on the page inside a `<b>` tag.

### DIFF
--- a/tests/_support/AcceptanceTester.php
+++ b/tests/_support/AcceptanceTester.php
@@ -20,7 +20,11 @@ class AcceptanceTester extends \Codeception\Actor
 {
     use _generated\AcceptanceTesterActions;
 
-   /**
-    * Define custom actions here
-    */
+	public function checkForPhpMessages() {
+		$I = $this;
+		$I->cantSee('Fatal error', 'b');
+		$I->cantSee('Warning', 'b');
+		$I->cantSee('Notice', 'b');
+	}
+
 }


veggiematts:
The test dit not fail after adding the checkForPhpMessages(); assertion:

```
run acceptance LoginCept.php --debug  --override="modules: config: WebDriver: url: http://coral.local"
[...]
 I am on page "/auth/"
  [GET] http://coral.local/auth/
 I check for php messages 
   I cant see "Fatal error","b"
   I cant see "Warning","b"
   I cant see "Notice","b"
 I fill field "loginID","coral_test"
 I fill field "password","coral_test"
 I click "#loginbutton"
 I check for php messages 
   I cant see "Fatal error","b"
   I cant see "Warning","b"
   I cant see "Notice","b"
 I see "Resources"
 I see "Licensing"
 PASSED 

```

t4k:
Are you working off the development branch? Do you have PHP messages set to display or are they suppressed?

jeffnm:
@t4k  do you still want this merged? It's not effecting any of the actual codebase, so I'm ok with merging it if you want, although we might be moving to a different acceptance test system if we continue to pursue the Travis CI solutions.

t4k:
I'm closing this because of the probability of moving to a new testing environment as @jeffnm mentions.

Issue #326, opened 2017-11-02
Non-static method called statically in Configuration.php


t4k:
Reported on the coral-user mailing list: http://lists.coral-erm.org/pipermail/coral-user/2017-November/000208.html

> **Deprecated**: Non-static method Utility::objectFromArray() should not be called statically in **…/auth/admin/classes/common/Configuration.php** on line **35**

Line 35 should probably read:
```php
$this->$section = (new Utility())->objectFromArray($entries);
```

~~Untested at this point.~~ This will have to be fixed in every module’s Configuration class.

t4k:
This does seem to be the solution. I will add a PR.

PR #327, opened 2017-11-02
Issue #326: Fix statically called non-static method.

t4k:
Addresses #326.

## Review Procedure
Enable PHP error reporting for E_DEPRECATED and load any CORAL page besides the root index and there will be Deprecated messages. Apply this fix and Deprecated messages for this issue are no longer there.

### DIFF
--- a/auth/admin/classes/common/Configuration.php
+++ b/auth/admin/classes/common/Configuration.php
@@ -32,7 +32,7 @@ class Configuration extends DynamicObject {
 
 		// Save config array content as Configuration properties
 		foreach ($config as $section => $entries) {
-			$this->$section = Utility::objectFromArray($entries);
+      $this->$section = (new Utility())->objectFromArray($entries);
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


remocrevo:
@t4k Watch out for whitespacing (tabs vs. spaces) as seen here:
https://patch-diff.githubusercontent.com/raw/coral-erm/coral/pull/327.diff

t4k:
My interpretation of the discussion at the end of #55 was that we would begin to use the new coding guidelines on all new patches, regardless of the state of the old code, and not update old code just for formatting issues, unless a separate formatting issue is being addressed.

t4k:
The new `.editorconfig` does solve the issue locally. It sets tab width to 2, which makes all the indents visually consistent. I’m not sure if we can set GitHub to respect those settings online though.

veggiematts:
Out of the space-tabs issue, the fix is ok. I would merge this.

veggiematts:
Moreover, the API is broken without this patch when error_reporting = E_ALL
This is a serious bug, I'm merging this.

PR #333, opened 2017-11-22
Fix autoload

veggiematts:
The use of php's __autoload was not recommended in the past, and it is now deprecated.

cf: http://php.net/manual/en/language.oop5.autoload.php
> Although the __autoload() function can also be used for autoloading classes and interfaces, it's preferred to use the spl_autoload_register() function. This is because it is a more flexible alternative (enabling for any number of autoloaders to be specified in the application, such as in third party libraries). For this reason, using __autoload() is discouraged and it may be deprecated in the future.

By experience, I've also seen that __autoload sometimes behave strangely.

This PR replaces all calls to __autoload with spl_autoload_register.

Moreover, it allows to remove all unnecessary explicit inclusions in the API.

### DIFF
--- a/auth/directory.php
+++ b/auth/directory.php
@@ -24,7 +24,7 @@ define('BASE_DIR', dirname(__FILE__) . '/');
 define('CLASSES_DIR', ADMIN_DIR . 'classes/');
 
 // Automatically load undefined classes from subdirectories of |CLASSES_DIR|.
-function __autoload( $className ) {
+spl_autoload_register(function ( $className ) {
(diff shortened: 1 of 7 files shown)


jsavell:
This PR (
https://github.com/coral-erm/coral/pull/299/files) also eliminates the use of `_autoload` while removing duplication across module directory files.

veggiematts:
Oh, right, I didn't see the removal of __autoload in PR#299.

veggiematts:
Fixed by #299

Issue #145, opened 2016-12-12
Parent/Child AJAX generates too many DB connections


remocrevo:
The following report comes from Josh Myles:

No luck with persistent database connections since I believe the code
is explicitly opening and closing connections. I tried prepending the
database server host name with "p:" in all configuration.ini files.

I ended up raising the MySQL connection limit (max_connections=10000)
as well as ulimits for the HTTPd user (nofile nproc both to 10000) and
the large resources loaded.

/resources/resource.php just needs to be refactored. The issue is just
a side effect of the AJAX way in which the page is built. It doesn't
scale for large resources.

The resource we've been testing has 1384 children, which results in
2785 database connections being opened when resource.php is loaded.

jeffnm:
I feel like someone made an attempt to fix this issue previously. I remember someone making changes to the number of mysql connections that were being opened. Perhaps that was a different module that was changed?

veggiematts:
I think the main problem is that a connection is opened for every object instantiated. More than the resources/resource.php page, it's the way we use Objects and database connections that needs to be changed imho.

veggiematts:
I did a small patch that turns the DBService class into a singleton, for the connection to be reused.

If you want to try it to see if it improves performances, it's here: https://github.com/biblibre/Coral/commit/18d017289846d1cf4f756a36cb1805cde64f2585

On my instance, it reduces the processing time for a resource with 3000 children from 5+ seconds to roughly 2 seconds.

veggiematts:
(Moreover, this patch would allow to momentarily change database, allowing us to use objects from other modules, instead of writing SQL statements to gather informations from other modules)

jeffnm:
I'm tagging this issue for 2.1.0 beta milestone. I think we should seriously try to get it in then. I suspect many longterm CORAL users have already changed their `max_connections` setting and so don't notice this as an issue, but the performance boost is enough of a reason to do it.

jeffnm:
Closing as #292 and #343 have both been merged.

PR #292, opened 2017-09-14
Add new routes to Coral API

veggiematts:
This PR adds new routes to the Coral API:

`GET /resources/@id`
Get resource by resource id
Example: http://coral.local/resources/api/resources/2

`GET /resources?identifier=@id`
Get resource(s) by ISBN or ISSN
Example: http://coral.local/resources/api/resources?identifier=12345

`GET /resources/@id/titles`
Get a package's titles
Example: http://coral.local/resources/api/resources/4/titles

`GET /resources/@id/licenses`
Get a resource's licenses
Example: http://coral.local/resources/api/resources/1/licenses

`GET /organizations/@id`
Get an organization
Example: http://coral.local/resources/api/organizations/536
This works whether the organization module is enabled or not

Please note that this PR includes the patch I proposed to fix Issue #145. This is not the patch proposed in PR #245, although it does the same thing.

I needed the patch for this pull request, because it allows to 
momentarily change Database, allowing us to use objects from other modules, instead of writing SQL statements to gather informations from other modules, which is way cleaner (needed for the Organizations route, in case the Organizations module is enabled).

PR #343, opened 2018-01-11
Issue 145 [Follow-up]: Use singleton for database connection

veggiematts:
As pointed out in #245 , the patch recently merged in development as part of #292 only addressed the problem for the resource module.

This PR extends the use of a single DB connection to all modules.

### DIFF
--- a/auth/admin/classes/common/DBService.php
+++ b/auth/admin/classes/common/DBService.php
@@ -22,6 +22,20 @@ class DBService extends Object {
 	protected $db;
 	protected $config;
 	protected $error;
+	private static $_instance; //The single instance
+
+
+/*
+	Get an instance of the Database
+	@return Instance
+	*/
+	public static function getInstance() {
+		if(!self::$_instance) { // If no instance then make one
+			self::$_instance = new self();
+		}
+		return self::$_instance;
+	}
+
 
 	protected function init(NamedArguments $arguments) {
 		parent::init($arguments);
--- a/auth/admin/classes/common/DatabaseObject.php
+++ b/auth/admin/classes/common/DatabaseObject.php
@@ -52,7 +52,7 @@ class DatabaseObject extends DynamicObject {
 		$this->primaryKeyName = $arguments->primaryKeyName;
 
 		$this->primaryKey = $arguments->primaryKey;
-		$this->db = new DBService;
+		$this->db = DBService::getInstance();
 		$this->defineRelationships();
 		//$this->defineAttributes();
 		$this->overridePrimaryKeyName();
--- a/licensing/admin/classes/common/DBService.php
+++ b/licensing/admin/classes/common/DBService.php
@@ -23,6 +23,20 @@ class DBService extends Object {
 	protected $db;
 	protected $config;
 	protected $error;
+	private static $_instance; //The single instance
+
+
+/*
+	Get an instance of the Database
+	@return Instance
+	*/
+	public static function getInstance() {
+		if(!self::$_instance) { // If no instance then make one
+			self::$_instance = new self();
+		}
+		return self::$_instance;
+	}
+
 
 	protected function init(NamedArguments $arguments) {
 		parent::init($arguments);
--- a/licensing/admin/classes/common/DatabaseObject.php
+++ b/licensing/admin/classes/common/DatabaseObject.php
@@ -52,7 +52,7 @@ class DatabaseObject extends DynamicObject {
 		$this->primaryKeyName = $arguments->primaryKeyName;
 
 		$this->primaryKey = $arguments->primaryKey;
-		$this->db = new DBService;
+		$this->db = DBService::getInstance();
 		$this->defineRelationships();
 		//$this->defineAttributes();
 		$this->overridePrimaryKeyName();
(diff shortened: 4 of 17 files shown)


jeffnm:
@fondrenlibrary would you review this PR? If it can be merged, then we can close #145 and #245

fondrenlibrary:
Got the following error when switching to this PR branch.
image

veggiematts:
Do you have the same error when switching to the development branch?
Are there any relevant error messages in the the log files?

fondrenlibrary:
Same error when switching to the development . No error spotted in the error log.

t4k:
@fondrenlibrary I have experienced this when testing other PRs that change the `common/configuration.ini` file:

```
[installation_details]
version = "2.0.0"
```

Make sure the version isn't set to 2.1.0; that's when I've seen this problem.

fondrenlibrary:
[Wed Jan 31 09:45:31.173891 2018] [:error] [pid 35085] [client 10.74.22.178:54606] PHP Fatal error:  Uncaught exception 'Exception' with message 'There was a problem with the database: Unknown column 'R.acquisitionTypeID' in 'on clause'' in /var/www/html/CoralMaster/resources/admin/classes/common/DBService.php:52\nStack trace:\n#0 /var/www/html/CoralMaster/resources/admin/classes/common/DBService.php(91): DBService->checkForError()\n#1 /var/www/html/CoralMaster/resources/admin/classes/domain/Resource.php(1228): DBService->processQuery('SELECT R.resour...', 'assoc')\n#2 /var/www/html/CoralMaster/resources/ajax_htmldata/getSearchResources.php(35): Resource->search(Array, 'R.createDate DE...', '0, 25')\n#3 /var/www/html/CoralMaster/resources/ajax_htmldata.php(24): include('/var/www/html/C...')\n#4 {main}\n  thrown in /var/www/html/CoralMaster/resources/admin/classes/common/DBService.php on line 52, referer: http://10.74.22.184/CoralMaster/resources/index.php

veggiematts:
Sounds like your database and codebase are out of sync.
  - If the code mentions R.acquisitionTypeID, then your database should be pre-multiple ordering:
 acquisitionTypeID is in the Resource table (currently: master)
  - If the code mentions RA.acquisitionTypeID, then your database should be post-multiple ordering:
 acquisitionTypeID is in the ResourceAcquisition table (currently: development)

This PR is based on master, and your database seems to be on development.
Therefore, you should not switch to this PR to test it, but rather merging it to the development branch.

fondrenlibrary:
Why this PR is based on the master branch instead of the development one?

veggiematts:
Another question could be: why would you switch branch instead of merging to the target branch? :)

That being said, seeing the previous commits on my branch, I was wrong to say that this PR is based on master, it's not, it's based on devel.
However, when this PR was made, multiple ordering was not already merged to devel, that's why it's not here.
That's one reason to always merge a PR on the target branch to test it (unless each PR is rebased everyday on the target branch, but no one does that, it's only done when conflicts arise)

The other reason to merge on the target branch instead of switching branch would be that you want to make sure that the development you're testing is correct towards the current state of the target branch, not the state of the branch at the time it started.

Does that make sense to you?

t4k:
If it helps anyone, I sometimes use a workflow that involves patching instead of all the merging in order to test PRs.

I checkout the current development branch, and then I apply a patch for the pull request. A quick one-liner would be:
`curl https://github.com/coral-erm/coral/pull/343.patch | git apply -v`
or
`wget -q -O - https://github.com/coral-erm/coral/pull/343.patch | git apply -v`

This way it is always easy to undo changes for a single pull request.

@veggiematts Do you have a recommended testing procedure? Which pages or forms would be ideal to check against this PR to look for problems?

fondrenlibrary:
After rebase the local biblibre-Issue_145_Followup_DB_Connections branch to the local development branch, the test can be done and I don't see any particular problems introduced by this PR. Thank @veggiematts and @t4k .

jcuenod:
Don't know if this helps but that error message can be found in `install/index.php`.

In short, I believe that error means that `index.php` has not been correctly configured for the upgrade. So it's it's expecting `$INSTALLATION_VERSION` to match the last element in the array of versions

    $INSTALLATION_VERSION = "2.1.0";
    $INSTALLATION_VERSIONS = ["1.9.0", "2.0.0", "2.1.0"];

The rationale for this is that when this is changed it means that whoever the maintainer is updating this file is confident that the upgrade path works from `2.0.0` to `2.1.0`.

veggiematts:
Thanks @fondrenlibrary 
@t4k I don't have any particular test plan for this, as it basically affects every database call made through an object. On a positive side, I guess that if something would be broken, everything would be broken.

fondrenlibrary:
@t4k , to get a peace of mind, I also tested with  our live license data, and can see this PR also solve the issue of displaying large child license documents which pr #245 is charged to solve.
image

t4k:
@veggiematts @fondrenlibrary 

Great! I'm confident in the testing from you two.

t4k:
I'll be merging this tomorrow unless there are objections from anyone before then.

PR #348, opened 2018-01-17
Fix Remove resource

jcraitz:
This is a fix for #244 to allow resource removal.  Currently on development if you try to remove a resource, you will get an error that there is no resourceID in the where clause.  This is triggered because a foreign key was updated in the big sql update, but not in the resource class for the function getContacts().

test plan:
1. from the resource module, click in to a resource
2. click the X to remove resource.

### DIFF
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -428,7 +428,7 @@ class Resource extends DatabaseObject {
 	public function getContacts() {
 
 		$query = "SELECT * FROM Contact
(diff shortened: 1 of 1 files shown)


veggiematts:
Thanks for noticing the problem, but this fix is not really a fix :)

Since contacts are now linked to orders, and orders are to be deleted when a resource is deleted, each order should delete its own contacts (note: I have the feeling that this was already done before, but it may have been lost in a merge / rebase).
See #349

PR #349, opened 2018-01-18
Remove orders on resource deletion

veggiematts:
Since contacts are now linked to orders, orders are to be deleted when a resource is deleted (note: I have the feeling that this was already done before, but it may have been lost in a merge / rebase).

Also fixes #348

### DIFF
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -1448,7 +1448,7 @@ class Resource extends DatabaseObject {
 
 		$query = "DELETE
 			FROM ResourceNote
-			WHERE resourceID = '" . $this->resourceID . "'";
+			WHERE entityID = '" . $this->resourceID . "'";
 
 		$result = $this->db->processQuery($query);
 	}
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -424,28 +424,6 @@ class Resource extends DatabaseObject {
 	}
 
 
-	//returns array of contact objects
-	public function getContacts() {
-
-		$query = "SELECT * FROM Contact
-					WHERE resourceID = '" . $this->resourceID . "'";
-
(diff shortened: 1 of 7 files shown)


jcraitz:
I saw at least two more functions in the Resource class that are now broken/not-needed.  Not sure if they've been refactored like you did here for contacts.
- removeResourceSteps()
- removeResourceNotes()

veggiematts:
removeResourceSteps isn't called anywhere in Coral (it was already the case in master) so it can be safely removed.

removeResourceNotes is indeed broken, I'll provide a fix.

veggiematts:
Resource contacts and notes are now properly removed on resource deletion.

veggiematts:
_Test plan:_

**For notes:**
 - create some notes on resource level (Product tab, in this caseResourceNote.entityID = Resource.resourceID)
 - create some notes on order level (Acquisitions, Access and Cataloging tabs, in this case ResourceNote.entityID = ResourceAcquisition.resourceAcquisitionID)
 - delete the resource
 - check in the database that all notes (both on resource and order level) have been deleted.

**For contacts:**
 - create contacts
 - archive some of them
 - delete the resource
 - check in the database that all contacts have been deleted, as well as ContactRoleProfiles values for these contacts

veggiematts:
This has to be merged before releasing the beta, to avoid database inconsistencies.

jeffnm:
The note and contact deleting worked as expected. However, when deleting a resource I received the following at the top of the page:
```
Warning: Invalid argument supplied for foreach() in /var/www/html/resources/admin/classes/domain/Resource.php on line 1408
Resource successfully deleted.
```

veggiematts:
Fixed.

veggiematts:
@jeffnm Did you have a chance to test this again with the last fix?

jeffnm:
@veggiematts sorry, I didn't get to it yet.

jeffnm:
I was able to test this and it seems to work well. Merging.

PR #345, opened 2018-01-12
Fix API for multiple orders on resource creation

veggiematts:
This PR allows #344 to work with #244 .
More specifically, a default order needs to be created when submitting a resource through the API, in order to be able to attach the acquisition type to it.

PR #352, opened 2018-01-24
fixes workflow start for new resources created by ajax form or API

jcraitz:
@veggiematts 
NOTE: This commit relies on PR #345.


veggiematts:
I've tested this PR in conjunction with PR#345 on devel, and it does fix the automatic workflow start on resource submission.

PR #362, opened 2018-02-08
fixes #361

### LINKED ISSUE #361: Empty neededBy date in API form results in submitting unix start date
I'll be submitting a PR for this immediately.

Right now, on development, if you submit the api_form with a blank neededBy date, it tries to convert that to a unix date for 0.

jcraitz:
probably not needed, but here's a testplan.
without this PR:
1. submit a new resource with the api_form, but don't fill in the neededBy date input. (it should be hidden unless you check the 'urgent' checkbox.)
2. look at the new resource record and you'll see an initial note with a neededBy date of 12/31/1969 or 01/01/1970.

with this PR:
2. you should not see a neededBy date at all in the initial note section.

### DIFF
--- a/resources/api_client/index.php
+++ b/resources/api_client/index.php
@@ -50,9 +50,11 @@ if (isset($_POST['submitProposeResourceForm'])) {
 //        addToDescriptionText($body,$descField);
 //    }
 
-    if(isset($body['neededByDate'])){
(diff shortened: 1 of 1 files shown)


veggiematts:
Tested OK.

jeffnm:
Looks good.

PR #314, opened 2017-10-20
Statistics and yearly costs dashboards

veggiematts:
This PR adds two dashboards: resource statistics and yearly costs. Both dashboards can be exported as CSV.

Data can be filtered by:
 - resource type, subject or acquisition type on resource level.
 - order type or cost details on acquisition level.

Data can be grouped by resource type, subject or acquisition type, with sub-totals.

dashboards_1

dashboards_2

PR #366, opened 2018-02-28
Fix dashboards for multiple ordering

veggiematts:
Dashboards (#314) are currently broken, since the merge of multiple ordering (#244). This PR fixes the SQL queries to be compatible with multiple ordering.

Test plan:
 - Add acquisitions to some resources
 - Check that the dashboards are not displayed, with SQL errors in the logs
 - Apply the patch
 - Check that the dashboards are now correctly displayed.

### DIFF
--- a/resources/admin/classes/common/Dashboard.php
+++ b/resources/admin/classes/common/Dashboard.php
@@ -16,9 +16,10 @@ class Dashboard {
 
         $query .= "
                  FROM Resource R
-                    LEFT JOIN ResourcePayment RP ON RP.resourceID = R.resourceID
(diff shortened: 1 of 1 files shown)


sbouis:
Hello,
I have tested this patch on a development branch. Before applying this patch, I have no result when I try after applying the patch, I have results in both dashboards : statistics and yearly costs. I have try several filters and several group by : displays and exports are right.

jeffnm:
Thanks @sbouis. Since you confirm it's working and since the changes only affect `resources/admin/classes/common/Dashboard.php` I'll go ahead and merge it.

PR #367, opened 2018-03-12
API: Fix licenses route

veggiematts:
The following route is currently broken on devel:
`resources/api/resources/@id/licenses`

> 500 Internal Server Error
> Invalid argument supplied for foreach() (2)
> #0 /home/coral/www/resources/api/index.php(327): flight\Engine->handleError(2, 'Invalid argumen...', '/home/coral/www...', 327, Array)

This PR fixes it.

It also enhance some existing routes:
 - Aliases are now embedded in `resources/api/resources/@id/` response.
 - Qualifiers are now embedded in `resources/api/resources/@id/licenses` response.

It also introduce a new route:
`resources/api/resources/@id/packages` to get the parent resources of a given resource.


jeffnm:
The code looked good to me, but I have not tested it.

t4k:
Tested with PHP 7.1 and this patch fixes the described issue.

PR #372, opened 2018-03-13
Add missing translation function

veggiematts:


scottvieira:
What should I say about this bug fix for the release notes?

veggiematts:
@scottvieira nothing: it's not worth mentioning in the release notes imho.

PR #373, opened 2018-03-13
Make Ebsco KB integration translatable

veggiematts:
PR #334 introduced EBSCO KB integration.
However, a lot of strings brought by this development do not use gettext and are therefore not translatable.
This PR fixes this.

Please note that a few strings are still not translatable, but this requires further investigation on how to make them translatable, and this is beyond the scope of this PR (titleSearchFieldOptions, titleResourceTypeOptions, selectionOptions, and packageContentTypeOptions in admin/classes/domain/EbscoKbService.php )

### DIFF
--- a/resources/ajax_htmldata/getEbscoKbTitleDetails.php
+++ b/resources/ajax_htmldata/getEbscoKbTitleDetails.php
@@ -31,24 +31,24 @@ $title = $ebscoKb->getTitle($titleId);
                 <div class="row">
                     <div class="col-6">
                         <dl>
-                            <dt>Publication Type</dt>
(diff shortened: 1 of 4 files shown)


jeffnm:
This looks pretty straightforward to me. I do have two questions/comments though. 

1. Could you open an issue about the remaining untranslatable strings? 
2. Should we create new label related to translation? Either "needs translation" or "translateable" or something like that?

veggiematts:
1. Yep: #377 
2. Yes, we should. I've created an "untranslatable" label to prevent PR's from being merged if there's no way to translate some strings.

PR #296, opened 2017-09-19
Coral / ILS interaction: vendors

veggiematts:
Background: see Issue #172 

This PR allows to synchronize vendors between Coral and an ILS.
In this PR, it has been done for Koha, but connectors to other ILSes can be made by implementing the following interface:

```
interface ILSClient {
    function addVendor($vendor);
    function getVendor($id);
    function getVendorByName($name);
    function getVendorByExactName($name);
    function vendorExists($name);
    function getILSName();
    function getILSURL();
    function getVendorURL();
}
```

To enable this feature, you have to add an [ils] section in organizations/admin/configuration.ini:

```
[ils]
ilsConnector=koha
ilsVendorRole="Vendor"
ilsApiUrl="http://mykoha.tld/api/"
ilsAdminUrl="http://pro.mykoha.tld/"
```

Once enabled, you can search for an existing vendor in Koha when creating an organization in Coral:
vendor_1

If the vendor does not already exist in Koha, it will be added if the vendor role has been selected.

If the vendor already exists in Koha, we can automatically retrieve its informations:
vendor_2
vendor_3

In both cases, the vendor can no longer be modified in Coral, as the ILS is authoritative.

Once the Coral organization and the Koha vendor are linked, a link to open the vendor in Koha is displayed:
vendor_4

PR #378, opened 2018-03-15
Allow to use OAuth for the ILS connector

veggiematts:
PR #296 introduced vendors synchronization with an ILS.
This PR adds OAuth support to authenticate against the ILS API.

See:
https://oauth.net/
http://oauth2-client.thephpleague.com/


jeffnm:
This pull request only changes Koha integration and OAuth, so I'm fine merging it without additional testing since Biblibre is supporting this feature and has tested it. I don't see anything in it that looks like it could impact other CORAL functionality.

PR #379, opened 2018-03-15
EBSCO Kb Integration: Warning about 10k records

queryluke:
## Testing

__Prerequisite__ have Ebsco API credentials so you can query EBSCO KB.

1. In the EBSCO KB search, find a package with over 10,000 records (Academic Search Complete is a good example). Click __Import__. You should have red warning under Package Titles.

2. Find a package with over 10,000 records and your library has < 10,000 records. The easiest way is to do a package search with Selection set to 'Not Selected'. Click __Import__. The same warning should appear beneath 'All Titles'. But Selected Titles and Package Only should still be an option.

Finally, if you want to edit the wording, it's at the top of the file in the `$ebsco10kLimitText` var

### DIFF
--- a/resources/ajax_forms/getEbscoKbPackageImportForm.php
+++ b/resources/ajax_forms/getEbscoKbPackageImportForm.php
@@ -185,8 +185,7 @@ if ($orgModule) {
                                     <div class="row">
                                         <div class="col-12 pb-1">
                                             <p>
-                                                You have <?php echo $package->selectedCount; ?> of <?php echo $package->titleCount; ?>
-                                                titles selected in EBSCO Kb.
+                                                <?php echo _('You have '.$package->selectedCount.' of '.$package->titleCount.' titles selected in EBSCO Kb.'); ?>
                                             </p>
                                         </div>
                                         <?php foreach($titleOptions as $option): ?>
--- a/resources/ajax_forms/getEbscoKbPackageImportForm.php
+++ b/resources/ajax_forms/getEbscoKbPackageImportForm.php
@@ -185,7 +185,7 @@ if ($orgModule) {
                                     <div class="row">
                                         <div class="col-12 pb-1">
                                             <p>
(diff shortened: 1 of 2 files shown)


resources/ajax_forms/getEbscoKbPackageImportForm.php
```diff
@@ -172,31 +177,41 @@
...
-                                        </label>
+                                <?php if ($package->selectedCount > 10000): ?>
+                                    <?php echo $ebsco10kLimitText; ?>
+                                    <input type="hidden" name="titleFilter" value="none">
+                                <?php else: ?>
+                                    <p id="span_error_titleFilter" class="smallDarkRedText"></p>
+                                    <div class="row">
+                                        <div class="col-12 pb-1">
+                                            <p>
+                                                You have <?php echo $package->selectedCount; ?> of <?php echo $package->titleCount; ?>
```
veggiematts: Please use gettext.

resources/ajax_forms/getEbscoKbPackageImportForm.php
```diff
@@ -185,8 +185,7 @@
                                     <div class="row">
                                         <div class="col-12 pb-1">
                                             <p>
-                                                You have <?php echo $package->selectedCount; ?> of <?php echo $package->titleCount; ?>
-                                                titles selected in EBSCO Kb.
+                                                <?php echo _('You have '.$package->selectedCount.' of '.$package->titleCount.' titles selected in EBSCO Kb.'); ?>
```
veggiematts: This won't behave nicely for translations (because the variables are in the translation string).

Something like this would be better:
`echo sprintf(_('You have %d of %d titles selected in EBSCO Kb'), $package->selectedCount, $package->titleCount);`

Or even better, to display singular or plural form, according to the number of titles:
`echo sprintf(ngettext('You have %d of %d title selected in EBSCO Kb', 'You have %d of %d titles selected in EBSCO Kb', $package->titleCount), $package->selectedCount, $package->titleCount);`

fondrenlibrary:
Good to go!

PR #390, opened 2018-03-22
Fix multiple issues with resource submission

veggiematts:
- Don't create multiple orders when a resource is saved as draft before being submitted
  Test plan:
  
   - Before the fix: Create a resource. Save it as draft, validate the resource. Two default orders will be created instead of one. If you edit the resource as draft multiple times before validating it, multiple default orders will be created.
 
    - After the fix: Proceed as before. Check that only one default order has been created when the resource is validated.

 - Fix acquisitionType selection when editing a resource saved as draft
   Test plan: 
     - Before the fix: Create a resource with an acquisitionType. Save it as draft. Edit it as draft. The acquisitionType will not be selected in the form.
     - After the fix: Proceed as before. Check that the acquisitionType is selected when editing the resource as draft. Check that the acquisitionType is registered in the default order when the resource is validated.
 
- Remove dead code: a ResourcePayment array was created and never used.

### DIFF
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -1395,6 +1395,14 @@ class Resource extends DatabaseObject {
 		$this->removeResource();
 	}
 
+    // Removes all resource acquisitions from this resource
+    public function removeResourceAcquisitions() {
+        $instance = new ResourceAcquisition();
+        foreach($this->getResourceAcquisitions() as $instance) {
+            $instance->removeResourceAcquisition();
+        }
+
+    }
 
 
 	//removes this resource
@@ -1404,12 +1412,7 @@ class Resource extends DatabaseObject {
 		$this->removeResourceOrganizations();
 		$this->removeAllSubjects();
 		$this->removeAllIsbnOrIssn();
-
-        $instance = new ResourceAcquisition();
-        foreach($this->getResourceAcquisitions() as $instance) {
-            $instance->removeResourceAcquisition();
-        }
-
+        $this->removeResourceAcquisitions();
 		$instance = new ExternalLogin();
 		foreach ($this->getExternalLogins() as $instance) {
 			$instance->delete();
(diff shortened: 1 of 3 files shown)


jeffnm:
This is the last open PR tagged for 3.0 with an important label. I'll try to take a look over the weekend and see if I can test it, unless someone else beats me to it, or it gets moved to the parking lot for this beta.

t4k:
I was getting a PHP Fatal Error when testing this:

> Fatal error: Uncaught Exception: There was a problem with the database: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'ORDER BY noteTypeID desc LIMIT 0,1' at line 4 in /var/www/coral/web/resources/admin/classes/common/DBService.php:52 Stack trace: #0 /var/www/coral/web/resources/admin/classes/common/DBService.php(91): DBService->checkForError() #1 /var/www/coral/web/resources/admin/classes/domain/Resource.php(373): DBService->processQuery('SELECT * FROM R...', 'assoc') #2 /var/www/coral/web/resources/admin/classes/common/Object.php(55): Resource->getInitialNote() #3 /var/www/coral/web/resources/ajax_forms/getNewResourceForm.php(54): Object->__get('getInitialNote') #4 /var/www/coral/web/resources/ajax_forms.php(24): include('/var/www/coral/...') #5 {main} thrown in /var/www/coral/web/resources/admin/classes/common/DBService.php on line 52

I'll commit a fix that adds quotes around the `nodeTypeID` in the SQL.

Otherwise this PR fixes the issues as laid out in the test plan.

t4k:
I'm probably the last one working today, so I'll go ahead and merge this, but anyone else can test it out after it is merged as well.

scottvieira:
@t4k Was this a bug introduced with the multiple orders feature?

veggiematts:
Yes, it was.

scottvieira:
@t4k and @veggiematts Thank you both!

PR #391, opened 2018-03-22
Allow to close a modal popup with the escape key

veggiematts:
Thickbox, which we use in Coral, and most of modal popup libraries, allow to close a modal popup with the escape key.
However, there is a bug in our modified version of thickbox that prevents this behaviour.

Test plan:
Before the fix: open a modal popup (new resource for example). You can't close the popup with the escape key.
After the fix: open a modal popup (new resource for example). You can close the popup with the escape key.

### DIFF
--- a/resources/js/plugins/thickbox.js
+++ b/resources/js/plugins/thickbox.js
@@ -34,7 +34,6 @@ function tb_init(domChunk){
 }
 
 function tb_show(caption, url, imageGroup) {//function called when the user clicks on a thickbox link
-
 	try {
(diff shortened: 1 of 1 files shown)


jeffnm:
Is there a reason this is only being applied to Resources?

jeffnm:
I'm going to go ahead and merge this as a bug fix for the resources module, and if you want to open the same PR with fixes for the other thickboxes in the other modules, that would be great, but no reason to hold this one up for that.

Issue #332, opened 2017-11-15
Installer Failed: $db_info empty


tmarnell:
I installed CORAL for my institution today and, at the end of the process, got stuck on this error:

"Database access was denied from @. Please ensure that you can access the database with the password you provided."

In install/providers/have_default_db_user.php, line 159 shows that the database username and host should have been printed on either side of the @. When I added die(var_dump($db_info)); within the switch statement, I found $db_info was null. The array was properly defined in line 136, but didn't make it into the anonymous function on line 142.

So in line 142, I added $db_info to the use construct, and then the installation completed without errors.

jeffnm:
@jcuenod I never noticed this issue when it was originally opened. Can you take a look at it? I see that `$db_info` is used on line 144, but not 142. Is there an edge case that this would fail on, but work for most installs?

veggiematts:
I can confirm this bug.
For me, it happened when doing so:

On the screen that says 

> "To begin with, we need a username and password to create the databases CORAL and its modules will be using."

I entered a non-admin mysql user.

The installer tries to revoke the coral user rights:
`providers/have_default_db_user.php, line 119:`
`$db->processQuery("REVOKE ALL ON {$db_details["dbname"]}.* FROM {$db_details["username"]}@{$db_details["host"]}");`

Since the user I entered did not have the rights to do so, the REVOKE failed, returning a 1045 error:

`$return->yield->messages[] = _("Database access was denied from {$db_info["username"]}@{$db_info["host"]}. Please ensure that you can access the database with the password you provided.");`

I can also confirm that `$db_info["username"]` and `$db_info["host"]` were empty.

jcuenod:
Yup, thanks @veggiematts! I guess this hasn't been picked up on is that people have been entering admin users pretty reliably.

I also notice that 142 has some parameters that I don't think are used. So I think it should be:

    "bundle" => function($version = 0) use ($config_files, $db_info) {

veggiematts:
By the way, maybe it's just me, but I think the installer is not saying clearly enough that an admin user is required.

jeffnm:
Thanks @veggiematts that's good to know. It's the kind of thing that once you've run it over and over again you stop even reading the notice because you know what is needed.

PR #392, opened 2018-03-22
#332 installer bug fix 

jeffnm:
This is supposed to fix the empty `$db_info` during the install process which resulted in error messages about '... denied from @ '  as reported #332 

It also attempts to improve the language used when asking for the user with database creation permissions to be more clear about what to expect. 

You can verify this by installing CORAL using a database user without permission to modify or create other users, so when it tries to create `coral_user` it will fail (silently, by the way, which is another issue I'll open shortly) and thus when it tries to use `coral_user` at the end of the process to actually run CORAL commands, you'll get 
screen shot 2018-03-22 at 2 45 44 pm

without this patch, and this 
screen shot 2018-03-22 at 2 39 33 pm
with the patch. 

There is also an installation SQL issue error that I'm seeing

`Duplicate column name 'ebscoKbID'
For statement: ALTER TABLE 'Organization' ADD 'ebscoKbID' INT(11)`

But if you refresh the page you'll proceed past it. 

I will open a separate issue to get the installation sql and scripts updated to the 3.0 stage and make sure that there isn't duplicate SQL in the install and the updates, which is what appears to be causing that particular issue.

### DIFF
--- a/install/providers/have_database_access.php
+++ b/install/providers/have_database_access.php
@@ -84,7 +84,9 @@ function register_have_database_access_provider()
 					require "install/templates/database_details_template.php";
 					switch ($version) {
 						case Installer::VERSION_STRING_INSTALL:
-							$instruction = _("To begin with, we need a username and password to create the databases CORAL and its modules will be using.")
+							$instruction = _("CREATE DATABASES")
+                    . "<br/>"
+                    .  _("To create the databases CORAL will use, we need a mysql username and password with permission to create databases. This username and password will not be stored.")
 										. "<br />"
 										.  _("If you would like to use pre-existing databases or custom database names. Use the advanced section to configure these settings.");
 							$db_access_vars = array_intersect_key($db_access_vars, ["username" => 1, "password" => 1, "host" => 1]);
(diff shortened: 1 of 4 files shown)


jcuenod:
Why does the instruction have `<h4>` and `<h5>` - shouldn't they be the same?

jeffnm:
The existing databases/custom databases instructions are a subsection of the Create Tables instructions as I read the page.

I wondered about using <h*> tags at all there since none of the other ones are (I don't think), but it was begging for some better headings.

jeffnm:
Also technically, there should be `<h2>` and `<h3>` elements somewhere on the page before using `<h4>` and `<h5>`, so I'm fine removing them and doing something else for the heading.

queryluke:
@jeffnm To fix `Duplicate column name 'ebscoKbID' For statement: ALTER TABLE 'Organization' ADD 'ebscoKbID' INT(11)`, you'll want to remove the following lines. It looks like when the ebsco KB branch was merged it reverted my changes to the `install.sql` files.

[resources/install/protected/install.sql](https://github.com/coral-erm/coral/blob/development/organizations/install/protected/install.sql#L115)

[organizations/install/protected/install.sql](https://github.com/coral-erm/coral/blob/development/organizations/install/protected/install.sql#L115)

jcuenod:
I was thinking that about the `<h*>` tags. You could use an `<b>` and maybe also a `<br />`. So I would definitely be in favour of an alternative :)

jeffnm:
I've fixed the header elements and the ebscoKbID column being added twice.

t4k:
See my test comment in #394. I believe this can be merged. We may want one more confirmation of it working for someone else.

PR #396, opened 2018-03-26
Fix CSV export

veggiematts:
CSV export is currently broken in devel. This patch fixes it.

### DIFF
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -1027,7 +1027,27 @@ class Resource extends DatabaseObject {
 									LEFT JOIN IsbnOrIssn I ON I.resourceID = R.resourceID
 									" . $licJoinAdd . "
 								" . $whereStatement . "
-								GROUP BY RPAY.resourcePaymentID
+                GROUP BY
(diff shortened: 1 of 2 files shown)


veggiematts:
The notice problem is fixed.
The Mysql 5.7 error seems to be the same as issue #235. However, I don't have a Mysql 5.7 instance right now to develop a fix.

t4k:
I got it working on MySQL 5.7 by adding columns to the GROUP BY clause.

The export is full with no notices now.

One sidenote, our export with 4429 records takes longer than 5 minutes (300 seconds in Apache) and required adjusting the Apache configuration. Is there a way to warn users about this or prevent timeouts? The solution/problem probably calls for a new issue…

jeffnm:
If it's primarily a MySQL 5.7 issue, I'll add that to the known issues for the Beta release and we can merge this.

t4k:
I added a commit to this PR that fixes the issue for MySQL 5.7 so it won't be an issue if this PR is merged.

jeffnm:
OK, in that case, I'll merge it. Thanks!

We can open a new issue for the export time out you found.

PR #397, opened 2018-03-27
Issue 94: Allow ftp:// organization urls

### LINKED ISSUE #94: Organizations FTP URL has malformed links
From an email on the listserv. I've verified that the ftp option writes an http url so the ftp://ftp.site.com is incorrectly written. 

> In the Organizations Module, I would like to add an ftp "shortcut" in the "login to url" space.
> 
> When I put in ftp://ftp.xxxxx.com
> 
> and use the arrow (link) to go there, I am directed to a local library page with a "page not found" error.
> 
> Is there any way to insert a shortcut to the ftp site (without copy & paste)?

queryluke:
Fixes #94 

## Fix
Instead of searching for a missing `http` string, instead search for a missing protocol `://`
Additionally, added CSS to allow urls to wrap. Very long urls would bleed over to the helpful links section

## Testing
In the organizations module, add an ftp url as the company url or the login url (login url under Accounts).

### DIFF
--- a/organizations/ajax_htmldata.php
+++ b/organizations/ajax_htmldata.php
@@ -230,7 +230,9 @@ switch ($_GET['action']) {
 		if ($organization->companyURL){ ?>
 			<tr>
 			<td style='vertical-align:top;text-align:left;width:140px;'><?php echo _("Company URL:");?></td>
-			<td style='width:320px;'><a href='<?php echo $companyURL; ?>' target='_blank'><?php echo $organization->companyURL; ?></a></td>
+			<td style='width:320px; word-break: break-word;'>
+                <a href='<?php echo $companyURL; ?>' target='_blank' style="text-transform: none;"><?php echo $organization->companyURL; ?></a>
+            </td>
 			</tr>
 		<?php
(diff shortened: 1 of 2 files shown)


organizations/ajax_htmldata.php
```diff
@@ -579,8 +581,8 @@ function generateDowntimeHTML($downtime) {
 				<?php if ($externalLogin['loginURL']) { ?>
 				<tr>
 				<td style='vertical-align:top;text-align:left;'><?php echo _("Login URL:");?></td>
-				<td><?php echo $externalLogin['loginURL'];
-					if (strpos($externalLogin['loginURL'], 'http') !== 0) {
+				<td style="word-break: break-all;"><?php echo $externalLogin['loginURL'];
+					if (strpos($externalLogin['loginURL'], '://') !== 0) {
```
veggiematts: This is wrong. We should add http:// in front of the URL if :// is not found. Should be:
`if (strpos($externalLogin['loginURL'], '://') === false) {`
(otherwise we end up with "http://ftp://")

PR #398, opened 2018-03-27
Error 404 when trying to delete non-existent resource

### LINKED ISSUE #93: Deleting a non existing resource gives a 200 instead of a 404
Using multiples times the deletion link of a resource gives the same result(status code + success message) although only the first request deleted it.
http://example.org/resources/ajax_processing.php?action=deleteResource&resourceID=89

queryluke:
Fixes #93 

## Fix
Add php response code and resource not found error message

## Test
1. Create a resource
2. Get the resource id
3. Ping the delete resource url: some-domain/resources/ajax_processing.php?action=deleteResource&resourceID=`resourceID`
4. Should receive a `Resource successfully deleted` message and 200 response code
5. Ping the same url
6. Should receive a `Resource not found. Error: <db error message>` and 404 response code

### DIFF
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -1399,9 +1399,6 @@ class Resource extends DatabaseObject {
 
 	//removes this resource
 	public function removeResource() {
-		if (empty($this->resourceID)) {
-			throw new Exception('Resource not found');
-		}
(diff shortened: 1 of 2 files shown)


veggiematts:
Tested ok. Code's ok. I'm merging this.

PR #399, opened 2018-03-28
Issue 347: Add an error message on the import tool page when mbstring is missing

### LINKED ISSUE #347: Resource import tool fails silently
The resource import tool fails silently if the PHP mbstring extension is not installed/enabled on the server (for Coral 2.0.1). It would be good if an error message or warning is provided to the user. Or at least that this requirement is detailed in the documentation for the tool (and/or the installation guide).

veggiematts:
Fixes #347

### DIFF
--- a/resources/import.php
+++ b/resources/import.php
@@ -917,6 +917,7 @@
 	}
 	else
 	{
+        if (extension_loaded('mbstring')) {
 ?>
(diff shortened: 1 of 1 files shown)


jeffnm:
This looks pretty straightforward to me. I don't see any reason not to merge it.

PR #400, opened 2018-03-29
(issue #388) German translation

veggiematts:
This patch adds a German translation to Coral and adds German as an option to the language pull down.

Test plan:
 - Apply the patch
 - Make sure you have the `de_DE.UTF-8 UTF-8` locale installed on your system (`sudo dpkg-reconfigure locales` on Debian-based systems)
 - Restart PHP and your webserver
 - German should be available in the language dropdown list, and Coral should switch to german when asked for.


xsong9:
@veggiematts Should I add this to the release note for 3.0 beta?

veggiematts:
@xsong9, yes!

PR #401, opened 2018-03-30
Issue 258: Fixes Importing issues

### LINKED ISSUE #258: Problems w/loading DB1(R4) reports in Usage Statistics
In the Usage Statistics module when loading DB1(R4) reports, either manually or via SUSHI, it appears that the User Activity data (Regular Searches, Searches - Federated and Automated, Result Clicks, Record Views) is not stored.

Even worse, when loading DB1(R4) reports via SUSHI, it appears that the only counts that CORAL stores is Record Views (and, as mentioned above, it does not appear to store the user activity text itself to go along with the counts themselves).

queryluke:
Fixes #258 

## Fixes
__Activity type was not stored__ 
Needed attribute added to the stats models ([here](https://github.com/coral-erm/coral/commit/9d6671871dbe9c82e03dde41102652575a3f962e) and [here](https://github.com/coral-erm/coral/commit/8ccb53302073a552b3da7f1526d094d3e23939a3?diff=split#diff-cd8ce9a2b8ea58fc17ca7faa102e675c))

__SUSHI imports for DB1 did not support Activity Type__,
__SUSHI import for JR1 incorrectly summed total__
Rewrote the parseXML part of the sushi service. 

### Import from file
1. Import a DB1 file. 
2. Activity type should be stored. (I'm not sure where this shows up in the interface, I have been checking directly in the Mysql DB, the `coral_usage.MonthlyUsageSummary` table)

__Example file found in #258__
__NOTE!!__ I have trouble importing single year sample files due to a different bug. The other bug is fixed in PR #328. I would advise merging that before testing this fix.

### Import from SUSHI
For each report type (JR1, BR1, and DB1)
1. Setup a sushi platform (can be done for all three at once)
2. click Run Now button for that platform (can be done for all three at once)
3. once run, click "View to Process"
4. Review stats
    * For JR1 & BR1, make sure the YTD values are correct (i.e., correctly summed)
    * For DB1, make sure the activity type column is filled in and there are multiple rows if multiple activity types exist for a single resource.
5. Click confirm and make sure the stats can be imported to the database.
6. For DB1 import, make sure activity type is saved in the database.


jeffnm:
The issues in #258 appear to be solved by this PR. I'm going to merge #328 and then this one.

Issue #394, opened 2018-03-22
Make sure the Installer/Upgrader SQL is ready for the 3.0 Beta Release


jeffnm:
There are some SQL commands that are being run twice, once in the install, and then again during the subsequent "update" step that happens automatically during the install. 

New SQL from this point forward should only go in the numbered releases with incremental file names per #302 and #285. Since we are releasing 3.0.0 instead of 2.1.0, we need to fix those version numbers in the files and directories that have already been merged into development.

If you find any SQL that is not being applied properly, please comment or edit this issue to add it.

- [ ] The install SQL for the ebscoKbID https://github.com/coral-erm/coral/blob/a375c474f1e8633c7a6afd530b7d5207096da800/organizations/install/protected/install.sql#L115 is the same as https://github.com/coral-erm/coral/blob/a375c474f1e8633c7a6afd530b7d5207096da800/organizations/install/protected/2.1.0/003-334.sql#L1 but it should only occur in the version upgrade as that is expected to be applied even in the case of a fresh install.
- [ ] `/module/install/protected/2.1.0` folders need to be renamed to 3.0.0
- [ ] The `organizations.php`  and other modules installation providers need their 2.1.0 cases renamed to 3.0.0

jeffnm:
#392 is now fixing the ebscoKbID column in both the install and update columns.

jeffnm:
#402 is now open for the version changes. 

We still need to verify that all needed database changes have been added correctly to the SQL, but all that have will be installed/upgraded.

veggiematts:
So #392 and #402 combined are the two PR needed to test this issue ?

jeffnm:
Yes, both are needed.

jeffnm:
I'd prefer not to merge them myself. Any chance you could merge them @t4k ?

t4k:
Merged!

PR #402, opened 2018-04-04
Issue 394 installer 3.0.0

jeffnm:
This pull request addresses the remaining elements of #394 that aren't addressed in #392 

It doesn't have the fixes from #392 included, but combined they should solve the problems.

We can't release the beta without this.


veggiematts:
The code seems ok.

t4k:
See my test comment in #394. I believe this can be merged. We may want one more confirmation of it working for someone else.

Issue #103, opened 2016-08-19
I18n: Language switch shows "English" when the UI is in french


tuxayo:
## Steps to reproduce:
- Get a browser configured in french (maybe setting `LANG=fr_FR.UTF-8; LC_COLLATE=fr_FR.UTF-8` in a terminal and launching the browser from there could work for people not having a french browser)
- Go in private browsing (this issue only happens without the language already saved in a cookie)
- Then you will see that the language switch shows "English" when the UI is in french.

Language switch on auth page:
![capture d ecran_2016-08-19_11-27-53](https://cloud.githubusercontent.com/assets/2678215/17806651/0c1f4482-6606-11e6-8977-7750deba72d7.png)
Language switch on homepage:
![capture d ecran_2016-08-19_12-13-26](https://cloud.githubusercontent.com/assets/2678215/17806702/5f3b375c-6606-11e6-9c5e-c2058bf9477e.png)
Language switch on other pages:
![capture d ecran_2016-08-19_12-13-14](https://cloud.githubusercontent.com/assets/2678215/17806707/62c5824c-6606-11e6-8aa8-1e1cddcda012.png)
## Problems
- the switch should show the currently set language
- not consistent with the behavior when the language is saved in the cookie. In that case, the switch shows the currently set language
- **to switch to english, one have to switch to french, then to english**

Issue #403, opened 2018-04-05
"Deutsch" displayed as default in select list


t4k:
When working off the current development branch, I have "Deutsch" as the selected option in the select list for the *Change language:* options throughout CORAL.

All the language strings are showing in English across all modules, so it is confusing that the select list is indicating German.

After selecting English it appears to get locked and sticks, but the initial value as Deutsch does seem to be incorrect.

Related to #388.

veggiematts:
Probably related or the same as issue #103

PR #404, opened 2018-04-06
Issues 103, 403: Fix selected language in dropdown when language cookie is not set.

veggiematts:
PR Not found :)

Fixes issues #103 and #403 

Test plan:
 - Clear your Coral language cookie
 - Without changing language (no cookie set), check that the browser's language is selected by default in language dropdown lists everywhere in Coral.
 - Change the language (language cookie is set), check that the language you chose is selected in language dropdown lists everywhere in Coral.

### DIFF
--- a/LangCodes.php
+++ b/LangCodes.php
@@ -26,5 +26,9 @@ class LangCodes{
         );
         return array_key_exists($code_lang, $name_lang) ? $name_lang[$code_lang] : "English";
     }
+
+    public function getBrowserLanguage() {
+        return str_replace('-', '_', substr($_SERVER["HTTP_ACCEPT_LANGUAGE"],0,5));
+    }
 }
 ?>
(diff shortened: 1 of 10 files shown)


t4k:
This works for me. I'll merge it.

scottvieira:
@t4k Was this a bug from 2.01 or a 3.0 beta fix?

t4k:
@scottvieira I'm not sure. I only noticed this testing 3.0 after the German translation was added. I would read over #103 and #403 to deduce.

veggiematts:
The bug was here since the introduction of translations. However, you didn't notice it before german was introduced, because english was the first option, alphabetically, and thus selected by default :)

But for french people for example, it was a bug from 2.0.1

scottvieira:
@veggiematts That's very helpful.  Thanks!

PR #416, opened 2018-04-13
Issue 412: Change email address to its plural form

veggiematts:
Before:
before

After:
after


LibClaire:
This works as expected.

Issue #423, opened 2018-05-17
Resource detail side menu not capitalized


jeffnm:
The side menu for looking at a resource record is not capitalized in English. This appears to be a bug that was masked by the old `text-transform: capitalize;` CSS.
![image](https://user-images.githubusercontent.com/651304/40183055-6b31ca7a-59b2-11e8-815b-1627ef2ed4f7.png)

PR #424, opened 2018-05-17
Capitalized resource_sidemenu links

jeffnm:
This is a bug fix for #423 

## Test Plan
Go to a resource record in English in the 3.0 beta release. The menu on the left will not be capitalized.

Apply this bug fix and reload the page. The menu on the left will be capitalized.

### DIFF
--- a/resources/directory.php
+++ b/resources/directory.php
@@ -84,16 +84,16 @@ function watchString($string) {
 function resource_sidemenu($selected_link = '') {
   global $user;
   $links = array(
-    'product',
-    'orders',
-    'acquisitions',
-    'access',
-    'cataloging',
-    'contacts',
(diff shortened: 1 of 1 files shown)


jeffnm:
This PR now includes capitalization for the same problem in the Usage module as well.

veggiematts:
Note: I'll have to modify and re-generate translations files before 3.0 if this is merged.

remocrevo:
@veggiematts Do you want to avoid remaking the translation files?  Or is that a normal part of a release now?  I can imagine fixing this bug by capitalizing the first letter of every tab programmatically instead of changing the labels themselves, but I think Jeff's change is better in the long term because it makes the code more clear.  So if you agree, I support merging this.

PR #425, opened 2018-05-18
Blank form in Resource Module and Fund List Dropdown not populating

fondrenlibrary:
Fix blank panes in Resource detail page found with Firefox (version 38.3 though very old)

Fix blank fund list  or newly added missing in fund list in Edit Cost From, for the Fund class' method  getUnArchivedFunds can only handle NULL value but not "0" of the archived field of the Fund table.
image
image


veggiematts:
Okay with the fix, although I wonder why my previous funds, like in your screenshot, had NULL, but the newly created ones have 0.

Issue #433, opened 2018-05-28
Some resource searches are broken


veggiematts:
The following resource search fields are currently broken:
  - Funds
  - Note type
  - Note content
  - Purchase site
  - Authorized site
  - Administering site
  - Authentication type
  - Cataloging status 
  - Workflow step

PR #434, opened 2018-05-28
Issue 433: Fix some resource searches

veggiematts:
Fix for issue #433 

This patch fixes the following resource searches:
Funds
Note type
Note content
Purchase site
Authorized site
Administering site
Authentication type
Cataloging status
Workflow step

### DIFF
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -699,19 +699,19 @@ class Resource extends DatabaseObject {
 
 
 		if ($search['authenticationTypeID'] == 'none') {
-			$whereAdd[] = "R.authenticationTypeID IS NULL";
+			$whereAdd[] = "RA.authenticationTypeID IS NULL";
 			$searchDisplay[] = _("Authentication Type: none");
 		}else if ($search['authenticationTypeID']) {
-			$whereAdd[] = "R.authenticationTypeID = '" . $resource->db->escapeString($search['authenticationTypeID']) . "'";
+			$whereAdd[] = "RA.authenticationTypeID = '" . $resource->db->escapeString($search['authenticationTypeID']) . "'";
 			$authenticationType = new AuthenticationType(new NamedArguments(array('primaryKey' => $search['authenticationTypeID'])));
 			$searchDisplay[] = _("Authentication Type: ") . $authenticationType->shortName;
 		}
 
 		if ($search['catalogingStatusID'] == 'none') {
-			$whereAdd[] = "(R.catalogingStatusID IS NULL)";
+			$whereAdd[] = "(RA.catalogingStatusID IS NULL)";
 			$searchDisplay[] = _("Cataloging Status: none");
 		} else if ($search['catalogingStatusID']) {
-			$whereAdd[] = "R.catalogingStatusID = '" . $resource->db->escapeString($search['catalogingStatusID']) . "'";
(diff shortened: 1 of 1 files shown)


queryluke:
@cjc66 Asked me to test this. Before it seems the searches were returning 500 errors. While they no longer return 500 errors, some of the searches do not appear to work as intended.

**Authorized Site, Administering Site**
Though I have an authorized and administering site set in the acquisitions tab, filtering on those values returns 0 results.
screen shot 2018-06-18 at 9 18 31 am

**Note Type & Content: Notes on the product tab are not search**
Though I have a Product Details note, filtering by Note Type "Product Details" returns 0 results.
screen shot 2018-06-18 at 9 25 52 am

However, if I filter by Note Type "Access Details", I get results because this second note exists in the acquisitions tab. The same is true for access & cataloging tabs.
screen shot 2018-06-18 at 9 30 31 am

resources/admin/classes/domain/Resource.php
```diff
@@ -815,12 +815,12 @@ public function searchQuery($whereAdd, $orderBy = '', $limit = '', $count = fals
...
-									LEFT JOIN ResourceAuthorizedSiteLink RAUSL ON R.resourceID = RAUSL.resourceID
-									LEFT JOIN ResourceAdministeringSiteLink RADSL ON R.resourceID = RADSL.resourceID
-									LEFT JOIN ResourcePayment RPAY ON R.resourceID = RPAY.resourceID
-									LEFT JOIN ResourceNote RN ON R.resourceID = RN.resourceID
-									LEFT JOIN ResourceStep RS ON R.resourceID = RS.resourceID
+									LEFT JOIN ResourcePurchaseSiteLink RPSL ON R.resourceID = RPSL.resourceAcquisitionID
+									LEFT JOIN ResourceAuthorizedSiteLink RAUSL ON R.resourceID = RAUSL.resourceAcquisitionID
+									LEFT JOIN ResourceAdministeringSiteLink RADSL ON R.resourceID = RADSL.resourceAcquisitionID
+									LEFT JOIN ResourcePayment RPAY ON RA.resourceAcquisitionID = RPAY.resourceAcquisitionID
+									LEFT JOIN ResourceNote RN ON RA.resourceAcquisitionID = RN.entityID
```
remocrevo: One problem is that the ResourceNote table "entityID" field only exists in the 3.0.0/002-350.sql upgrade script, so it (and maybe more) needs to be added to the main install.sql (right? is that still our approach to fresh installations?).

veggiematts: I'm not sure. I thought we decided that fresh installs would process install.sql *and* the upgrade scripts. Can someone confirm this?

veggiematts:
This follow-up should fix the problems @queryluke mentioned.

resources/admin/classes/domain/Resource.php
queryluke: @veggiematts, correct. When installing the scripts process `install.sql` and all upgrade sql files as long as the upgrade files are contained within a version directory, e.g. `/$MODULE/install/protected/$VERSION`

Here is the [PR](https://github.com/coral-erm/coral/pull/354)

queryluke:
Yes, I have confirmed this is working.

resources/admin/classes/domain/Resource.php
remocrevo: @queryluke @veggiematts Got it. Thanks for the clarification.

PR #435, opened 2018-06-13
Add ImportHistory table update

veggiematts:
ImportHistory table creation is neither in install.sql nor in a file processed by the updater.
This patch allows the ImportHistory table to be created (needed for #170 )


veggiematts:
Needed since PR #354

PR #436, opened 2018-06-14
Add an option to save import configuration on the fly

veggiematts:
This patch allows to save an import configuration on the import page, allowing to create/edit configurations during the import process.

savemapping1
mapping2
mapping3

Also, bug fix: currency, orderType and sendemails attributes were not properly
transmitted through import process

Test plans:
Create a new config:
 - Start an import
 - Create a configuration
 - Save current configuration
 - Check in admin -> Import Configuration that the configuration has been added

Edit an existing configuration and save it as new:
 - Start an import
 - Load an existing configuration
 - Edit it
 - Save it with a new name
 - Check in admin -> Import Configuration that the configuration has been added

### DIFF
--- a/resources/css/style.css
+++ b/resources/css/style.css
@@ -2814,6 +2814,19 @@ img.remove {
     padding-bottom: 5px;
 }
 
+#saveCurrentMappingDiv fieldset legend {
+	padding: 0 5px;
+	font-weight: bold;
+	width: initial;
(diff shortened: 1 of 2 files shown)


LibClaire:
I have tested this and it works as expected. I was able to create and edit new Import configurations during the import process.

Issue #439, opened 2018-06-19
Resource accounts tab is not displayed anymore


veggiematts:
Clicking on "Accounts" in the resource menu makes the menu and the tab content disappear. Caused by PR #424 ?

jeffnm:
I can reproduce the issue, but interestingly if I change the capitalization back to 'accounts' in the `directory.php` file, the accounts menu entry disappears. I think the Accounts menu entry is actually not needed for Resources. It's not displayed in 2.0.1 on our production server. 

Does anyone know what the Accounts menu item was originally intended for? If it's not needed here, we can remove it.

jeffnm:
Two alternative solutions: #441 Remove the Accounts menu entry or #443 Restore the Accounts menu entry.

When one is selected and merged, we can close this PR, merge #442 translations, and then release 3.0.

techsvcslib:
It took me a little while to understand what is going on here. If I understand correctly, this refers to the accounts vertical tab when looking at an individual resource record. To confirm: this is indeed important functionality and must be fixed.

jeffnm:
Merged #443

PR #441, opened 2018-06-20
Removed "Accounts" entry from Resources menu

jeffnm:
Since I don't see the Accounts menu in 2.0.1 or when it's lowercase, I'm proposing we solve #439 by removing that entry from the menu.

### DIFF
--- a/resources/directory.php
+++ b/resources/directory.php
@@ -90,7 +90,6 @@ function resource_sidemenu($selected_link = '') {
     'Access',
     'Cataloging',
     'Contacts',
-    'Accounts',
     'Issues',
     'Attachments',
     'Workflow',


veggiematts:
Okay, but if accounts are really not used anymore, we should do more cleanup:

 * function updateAccounts in js/resource.js
 * ajax_htmldata/getAccountDetails.php
 * ajax_forms/getAccountForm.php
 * ajax_processing/submitExternalLogin.php
 * admin/classes/domain/ExternalLoginType.php
 * admin/classes/domain/ExternalLogin.php
 * js/forms/externalLoginForm.js
 * ExternalLogin table
 * Update the documentation (See Account in http://docs.coral-erm.org/en/latest/resources.html)

On the other hand, if accounts are still used/needed, see PR #443

jeffnm:
#443 is merged instead.

PR #443, opened 2018-06-21
Issue 439: Restore accounts tab

veggiematts:
Restore the accounts tab in the resource module, see Issue #439 
Alternative resolution, see PR #441


jeffnm:
Tested, it works. This would get my vote over #441 for the 3.0.0 release.

PR #444, opened 2018-06-25
Make resource and usage sidemenus translatable

veggiematts:
The resource and usage sidemenus are currently untranslatable. This patch makes them translatable.
Also, it removes icons and code that were unused since the last graphical revamp.


poacherone:
Tested, and seems to work as intended. Note that I can't see if the menus are now actually translatable, but this PR doesn't brake anything as far as I can tell.

jeffnm:
Translation was partially supported, but it's better to translate them in the array rather than when echoing the menu name out, as that allows for the same translated string to be used for the title and the text. 

The removal of loaded but hidden icons from the usage module is also an improvement.

jeffnm:
I'm not sure what's going on, but with this new code, the Turkish translation loses its icon for the Import menu in the Usage module, and fails to load the actual record as well. 
New version:
screen shot 2018-10-24 at 9 09 08 am
Old version:
screen shot 2018-10-24 at 9 10 01 am

French, English, and German all load correctly, as do the Chinese translations, so I'm guessing this is a bug in the Turkish translation. What do you think @veggiematts?

jeffnm:
Great! That works.

Issue #448, opened 2018-07-02
Can't switch language if Coral is running on a non standard port


kfischer:
When running Coral on a non-standard port, for example :9090, switching the language from English to another is not possible. When choosing another language from the pull down, the page reloads and jumps back to English instead of the language chosen.

This is because the lang cookie storing the selected language isn't set.

After some research the problem is that Coral tries to set the domain attribute of the cookie to the host including the port number. Instead it should be using the host name without the port.

Cookies do not provide isolation by port. If a cookie is readable by  a service running on one port, the cookie is also readable by a service running on another port of the same server... (https://tools.ietf.org/html/rfc6265#section-5.1.3)

Test plan:
- Configure Coral to run on another port, like 9090
- Try to switch to another language than English from different modules and pages
- Check that no lang cookie is stored 
- Apply the patch
- Repeat the test plan and verify that the cookie is now set properly and switching the language works as expected

kfischer:
I am going to provide a pull request for this.

PR #449, opened 2018-07-02
(issue #448) Can't switch language if Coral runs on a non-standard port

kfischer:
When running Coral on a non-standard port, for example :9090,
switching the language from English to another is not possible.
When choosing another language from the pull down, the page
reloads and jumps back to English instead of the language chosen.

This is because the lang cookie storing the selected language isn't set.

After some research the problem is that Coral tries to set the
domain attribute of the cookie to the host including the port
number. Instead it should be using the host name without the port.

Cookies do not provide isolation by port. If a cookie is readable
by a service running on one port, the cookie is also readable
by a service running on another port of the same server...
(https://tools.ietf.org/html/rfc6265#section-5.1.3)

Test plan:
- Configure Coral to run on another port, like 9090
- Try to switch to another language than English from different modules and pages
- Check that no lang cookie is stored
- Apply the patch
- Repeat the test plan and verify that the cookie is now
  set properly and switching the language works as expected

### DIFF
--- a/auth/admin.php
+++ b/auth/admin.php
@@ -158,7 +158,7 @@ if (isset($user) && ($user->isAdmin) && ($user->getOpenSession())){
         var cookievalid=2592000000; // 30 days (1000*60*60*24*30)
         time += cookievalid;
         now.setTime(time);
-        document.cookie ='lang='+lang+';path=/'+';domain='+wl.host+';expires='+now;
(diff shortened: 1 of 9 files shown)


veggiematts:
Can confirm switching language doesn't work on non-standard port without the patch.
Can confirm switching language does work on standard and non-standard port with the patch.
The code looks ok.
I'm in favor of merging this.

kfischer:
Thanks a lot for testing!

Issue #450, opened 2018-07-03
Can't update or delete accounts


sbouis:
When we add an account on a resource with a autorized login, we don’t find any link to update it or to delete it.
![image](https://user-images.githubusercontent.com/31436815/42211120-0f6174a4-7eb3-11e8-9f86-41f3c6cce6f9.png)

jeffnm:
This looks like it's an issue for the Contacts tab as well, where new contacts can be added but not edited or deleted.

veggiematts:
@jeffnm I don't have any problems editing/deleting contacts on the development branch. Can you provide more informations?

jeffnm:
@veggiematts my apologies, I was in a hurry when checking to see if this was a known issue and didn't notice that you had a pull request open for it already. I'll add the pull request to the patch 2.0.1 project as well. Thank you!

jeffnm:
A fix has been merged and will be included in the next release.

PR #452, opened 2018-07-04
Issue 450: Fix edition/deletion of local account 

veggiematts:
Issue #450 : Allow edition/deletion of local account 
(ie: not an account inherited from the organization module)


LibClaire:
I tested this and was able to add, edit and delete accounts. Works as expected.

I could not reproduce the problem from issue #450  which prevented deleting or editing contacts. I was able to edit and delete these also (with and without this fix applied).

jeffnm:
I also tested it, and was able to reproduce #450 and this pull request allowed editing and deleting non-inherited accounts.

PR #453, opened 2018-07-11
Restore workflow archiving fix

jsavell:
Restores PR #213 fix for Issue #211 

I enabled EditorConfig in my text editor, and it looks like it went after some of the whitespace/line endings, but the only functional change is inside the isCurrentWorkflowComplete() method.


poacherone:
Tested this PR, and it seems to work as intended - an entire workflow can now be marked as complete again. We noticed a related issue though: [issue 497](https://github.com/coral-erm/coral/issues/497)

Issue #384, opened 2018-03-16
contact display improvement


PaulPoulain:
the contact list displays contacts one by one. It should be a table instead
![image](https://user-images.githubusercontent.com/4696271/37516777-97ceb22e-290f-11e8-8048-b04a29bccad1.png)

should be something like:

Name | Title | Role | Phone | Fax | email
---- | ---- | ---- | ---- | ---- | ----
tt | tt | Accounting | tt | tt | tt
yy | yy | support, general | yy | yy | yy

PaulPoulain:
linked to #380

PR #459, opened 2018-07-19
Improve contacts and attachments display

veggiematts:
This PR aims to improve contacts and attachments display in a more readable, compact way.

**Contacts:**
_Before:_
contacts_before
_After:_
contacts_improved

**Attachments:**
_Before:_
attachments_old
_After:_
attachments_improved

Note that the filename has been added in the attachments view.

### DIFF
--- a/resources/admin/classes/domain/ResourceAcquisition.php
+++ b/resources/admin/classes/domain/ResourceAcquisition.php
@@ -693,32 +693,34 @@ class ResourceAcquisition extends DatabaseObject {
 
 
 	//returns array of contact objects
-	public function getArchivedContacts() {
+	public function getArchivedContacts($moduleFilter=false) {
 
 		$config = new Configuration;
 		$contactsArray = array();
 
 		//get resource specific contacts
-		$query = "SELECT C.*, GROUP_CONCAT(CR.shortName SEPARATOR '<br /> ') contactRoles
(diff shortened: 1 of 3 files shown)


ehartnett:
I’ve personally never had a problem reading the contacts but if you’re going to make the change, I would not concatenate the title on the front of the person’s name. Others can chime in as they may do things differently but we do not use that field for “Mr., Ms., etc.”, we use it for that person’s job title, e.g., “Sales Associate, Account Manager, etc.” I wouldn’t want that stuck on the front  of their name.

One change to the contacts that would be nice would be to be able to prioritize them in some way. For larger companies there could be 5+ contacts but one of them is the primary. Since they’re sorted by first name, the primary person could be on the bottom. Our work around has been to add an asterisk to the front of their name so they display at the top.

-Eric

scottvieira:
To be honest, I prefer the current way for reading the contacts.  Instead, I would like to see other improvements being made to organizations.  Either way, I think these types of changes should be discussed by the steering committee first.  I agree also with Eric that the file name shouldn't come after the name.  I would put it at the end.

veggiematts:
> I would not concatenate the title on the front of the person’s name
done

> One change to the contacts that would be nice would be to be able to prioritize them in some way.
That would be nice, but beyond the scope of this development, imho

 > the file name shouldn't come after the name. I would put it at the end.
done

This development is linked to #384

Issue #462, opened 2018-07-27
Issue with Access Information in CORAL 3.0 Resources Module


rtatterson:
There appears to be a bug in the Resources module with editing access information. Data not being saved in the Access Information box when using "edit resource" but will save if using "edit access information."

jeffnm:
It appears that the pencil and paper icon on the Access Information box is not passing in the resourceAcquisitionID, just the resourceID. We'll need to update that ajax Access Form request to match the "edit access information" link.

jeffnm:
A fix has been merged and will be included in the next release.

PR #466, opened 2018-08-17
Fix access edition when using icon link

veggiematts:
Fixes issue #462


LibClaire:
I have tested this, access information saves correctly when using the edit icon.

PR #467, opened 2018-08-22
Add PIG (Public Interface Generator) addon into main Coral repository

veggiematts:
This PR integrates the PIG addon to the main Coral codebase.
PIG is originally available here: https://github.com/TAMULib/CORAL_Addons/tree/master/resources/pig

No database update is needed.

We may want to add a way to enable/disable PIG though.


veggiematts:
Once the patch applied, PIG is available at this address: /resources/pig.php

LibClaire:
I applied the patch and was able to navigate to /resources/pig.php, view resources and link out to them.

jeffnm:
I also was able to access the PIG without authentication as expected.

I agree we should definitely have a toggle for PIG and I think it should be off by default. 

I'm not sure if we should add PIG to the code base without that option since it's always been an Add-on and not a default functionality in the past. Let's get some more feedback on this.

xsong9:
I would like to keep it as an add-on, not a default functionality, simply because not all users would use it. I also like the idea of having a toggle to make the switch easier for users to use.

veggiematts:
Whatever's decided, here's a patch providing a configuration entry for PIG, disabled by default.

My opinion on this: we're currently trying to make things more integrated (we've already merged the git repositories in one, we were talking about merging the databases in one, etc.) and easier to install. 
I think it wouldn't hurt to provide PIG available out of the box, without the hassle of installing it from an external repository.
That being said, I totally agree with the fact that it should be disabled by default.

poacherone:
I agree that it makes sense to add PIG to the codebase, but have it configurable and off as default.

jeffnm:
I think your on/off option is good enough for an initial inclusion. I agree that it's better to keep all the CORAL code in one repository, and give easy options to turn things on or off. 

Is there a reason you called the configuration entry "pIG" rather than "PIG"? You were consistent, so it works, but I'm curious.

veggiematts:
Can we merge this?

PR #469, opened 2018-08-22
Fix access method deletion in admin

veggiematts:
Test plan:

1) On master or development branch, try to delete an unused access method in admin. The access method cannot be deleted.
2) Apply the PR
3) Try to delete an unused access method in admin. The unused access method can be deleted. Check that a used access method cannot be deleted.


jajm:
Hi,

I followed the test plan and the patch works as expected.

PR #488, opened 2018-09-22
Adds Terms tool code into Coral

queryluke:
## To test

1. Navigate to the licensing module and click on Admin
2. At the bottom of the page, there is a new setting for the terms tool which allows you to choose a service (SerialsSolutions, SFX, or EBSCO).
3. After selecting the service, enter the required variables for your service.

### Setting up a license
(skip this if you already have licenses & resources made)
1. In the Resources Module, create a new resource (it should have an ISBN or ISSN)
2. In the Licensing Module, create a new license
3. Select the new license, and "upload a new document" (you can provide any values but do not select Archived, you may upload any document)
4. On the license page, select Expressions on the left and "add new expression", for __Expression Type__ you need to select a type that is setup for terms tool display. _Interlibrary Loan_ and _Course Packs_ is setup by default. To make a new one (or edit and existing one), you may do so in the Admin utility. Either add or update an expression type with a _Note type_ of "Display." To get the check mark or X mark to show up in the terms tool, you also need to setup qualifiers of either Permitted or Prohibited for the expression (in the admin tool)
5. After adding an expression to the license, you must also check beside the expression name: 
screen shot 2018-09-22 at 1 59 22 pm
You may optionally add a display note.

6. On the license page, select Terms Tool and "add new terms tool resource link"

7. Choose your document and enter the resource. __Note:__ the string entered in Resource must _exactly_ match the string returned from the Terms Tool Service. 
screen shot 2018-09-22 at 2 06 42 pm

For example, I am using the EBSCO service, so the name must match the package name that the resource comes from. In the above example that name is "Academic Search Complete"

### Test that the terms tool works

To test the resource, provide the resource's ISxN to the terms tool route:
`http://your.coral.domain/licensing/terms.php?issn=xxxx-xxxx`
or 
`http://your.coral.domain/licensing/terms.php?isbn=xxxxxxxxxxxxx`

Here is an example output
screen shot 2018-09-22 at 2 13 01 pm

### Note about terms route
You may notice there is a change to the terms tool route. In older versions of coral it used to be `/terms?issn=xxxx-xxxx`. But users may have there older version of terms tool installed in this directory. So instead, I created a new file in the licensing module and moved the necessary files. Thus, any installation that upgrades their code base won't have an issue with conflicting files.

### Terms tool CSS [edit: 2020/02/15]
Per the request of Sirsidynix, I've added their updated css for the terms two. The base styles are found in `licensing/css/base-terms.css`. If a site would like to update the colors or adjust the css, they can do so by creating a `licensing/css/custom.css` file.


andyp-uk:
Have tested this on several systems in house and appears to be working as intended with EBSCO HoldingsIQ.

PR #495, opened 2018-10-03
OAuth authentication: bugfix and enhancements

veggiematts:
This patch provides a bugfix and several enhancements for OAuth authentication in the Koha connector:
  - Fix incorrect use of "Authorization" header.
  - Add better error handling
  - Authentication token is now stored in session and re-used until expired.
  - Replace incorrect call to Koha API.

Test plan:
Koha version >= 18.05 needed.
  - Coral: Install Unirest and league/oauth2-client in resources/api_client (composer might be the easiest way to do it)
  - Koha: Install Net::OAuth2::AuthorizationServer and enable RESTOAuth2ClientCredentials syspref.
  - Koha: Create a user for the REST API, edit this user, click on more -> manage API Keys -> Generate new client id/secret pair
  - Coral: edit organizations/admin/configuration.ini as such:
```
[ils]
ilsConnector=koha
ilsVendorRole="Vendor"
ilsApiUrl="http://yourkoha.tld/api/v1"
ilsAdminUrl="http://admin.yourkoha.tld/"
oauthid="<paste Client ID generated in Koha>"
oauthsecret="<paste Secret generated in Koha>"

```
  - Check that you can synchronize vendors between Coral and Koha (see http://docs.coral-erm.org/en/latest/integration.html and/or https://github.com/coral-erm/coral/pull/296 for explanations and screenshots) or that you get an error message if something went wrong.

### DIFF
--- a/organizations/admin/classes/api_client/KohaClient.php
+++ b/organizations/admin/classes/api_client/KohaClient.php
@@ -31,7 +31,7 @@ class KohaClient implements ILSClient {
             $oauth = new OAuth();
             $token = $oauth->getToken();
             if ($token) {
-                $headers['Authorization: Bearer '] = $token;
+                $headers['Authorization'] = 'Bearer ' . $token->getToken();
             }
         }
         return $headers;
@@ -80,8 +80,10 @@ class KohaClient implements ILSClient {
      */
     function vendorExists($name) {
         $headers = $this->authenticate();
-        $response = Unirest\Request::get($this->api . "/acquisitions/vendors/?name=$name");
-        return (count((array) $response->body) > 0) ? true : false;
+        $response = Unirest\Request::get($this->api . "/acquisitions/vendors/?name=$name", $headers);
+        $error = $this->_checkForError($response);
+        if ($error) return $error;
+        return (count((array) $response->body) > 0) ? 1 : 0;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


jeffnm:
@veggiematts We aren't Koha users, so I can't really test this well. The code looks good to me, but is there anyone else you could get to test this who uses Koha and Coral? 

Since nothing in these changes should cause any issues for non-koha users (no one else is using the ils api, right?), I am happy to accept Biblibre testing, but it would be nice to have another user if possible.

Also, do you think we should add installer support for setting those [ils] configurations?

veggiematts:
> is there anyone else you could get to test this who uses Koha and Coral?

I'm trying to :)

> no one else is using the ils api, right?

That is correct.

> Also, do you think we should add installer support for setting those [ils] configurations?

I think that would be a good thing, but not top-priority (as using the command line is mandatory anyway for installing dependencies with composer)

LibClaire:
I have tested this with OAuth Authentication.

I was able to create vendors (Organizations) in Coral which appeared in my Koha Vendors and was able to pull Koha vendor information into new organization in Coral. Worked as expected.

jeffnm:
Thanks @LibClaire and @veggiematts, that's good enough for me. I'll merge.

Issue #484, opened 2018-09-13
Hard-coded dates in SushiService


vinny75:
Was running into a problem where SUSHI was only pulling Jan-2018 through Feb-2018 despite setting the dates to something else.  Think I have tracked the problem down to these two lines in SushiService.php from commit 8ccb53302073a552b3da7f1526d094d3e23939a3

614 $startDateArr = explode("-", '2018-01-01');
615 $endDateArr = explode("-", '2018-02-28');

Changing these two lines fixed the problem for me

614 $startDateArr = explode("-", '$this->startDate);
615 $endDateArr = explode("-", $this->endDate);

LibClaire:
I can confirm that this worked as a fix for us too. Our SUSHI stats are now good!

veggiematts:
@queryluke Was there a reason you put fixed dates in your commit or was this a mistake? (I would go for the mistake, but still asking to be sure)

queryluke:
Yes, that seems like a mistake. Sorry about that. I probably hardcoded for testing and forgot to switch it back.

veggiematts:
Okay, thanks.

jeffnm:
A fix has been merged and will be included in the next release.

PR #496, opened 2018-10-09
Fix hard-coded dates in SushiService

veggiematts:
Fixes issue #484


jeffnm:
Based on the discussion in #484 and the straight-forward nature of this change, I am comfortable merging this.

Issue #489, opened 2018-09-28
Subscription dates not set correctly for orders during upgrade


poacherone:
When upgrading to Coral 3.0.0 from 2.0.1 the upgrade script doesn’t migrate all the Acquisitions data to the new ResourceAcquisitions table. The columns currentStartDate and currentEndDate are kept in the Resource table while all newly created orders are instead given today’s date in both subscriptionStartDate and subscriptionEndDate (“NOW()” on line 100 in the [upgrade SQL](https://github.com/coral-erm/coral/blob/master/resources/install/protected/3.0.0/002-350.sql)).

As far as we can tell there is no way to access/edit currentStartDate or currentEndDate from the UI in 3.0.0., and the newly created orders have incorrect dates. We believe this to be a bug. 

Preferably the dates from currentStartDate/EndDate should be migrated like the other Acquisition/Order data during the upgrade (to subscriptionStartDate/EndDate), and then drop the columns from Resource.

veggiematts:
After doing some digging, it feels like a bug to me too.
However, I'd like a second opinion on this.
In any case, if it is a bug, it should be fixed for 3.0.1

veggiematts:
Proposal:
  - update each first resourceAcquisition (ie: the one created by default during the 3.0 upgrade) with the dates of the matching resource
  - remove all remaining occurrences of the resource dates in the code.
  - drop the dates columns for the resource table.

poacherone:
I agree with @veggiematts proposal, it makes sense. This issue would really need to be fixed for Coral 3.0.1 I think, since ALL orders will have the wrong subscription dates without it.

jeffnm:
@veggiematts your proposal sounds good to me as well. The Resource table's currentStartDate and currentEndDate are currently still present in a 3.0.0 fresh installation as well, since we opted to change the install method to run the upgrades in sequence rather than updating the install.sql file every time we do a new release.

veggiematts:
There are two problems with this proposal though:
  1. I'm not sure how to update only the first ResourceAcquisition with a SQL query only (and we cannot use php scripts to update the database)
  2. If some users did modify the default order, their modification will be discarded.

jeffnm:
I think we could update the ResourceAcquisitions by looking for Resource Dates that are not null with a matching resourceID. Anything that was created afterwards will have null in those values.  

[...]

Issue #451, opened 2018-07-03
Default orders sort from newest to oldest


sbouis:
Hi,
I suggest that it would be easier for users if orders are sorted by date from newest to oldest.
When we have several orders, we often want to work on the more recent ones.

veggiematts:
Hi,

Currently orders are displayed in the order select dropdown list in no particular order, meaning it will most likely be in the order they have been created (but not necessarily, that's up to the SQL server).

It would be a good idea to define an order in which we should display them.

But what do you mean by "sorted by date from newest to oldest"?

  - sorted by reverse order of their creation?
  - sorted by reverse of their subscription start date?
  - sorted by reverse of their subscription end date?

poacherone:
I would say the last option: sorted by reverse of their subscription end date.

sbouis:
What could be marvellous : first sorted by reverse of their subscription end date and then secondly sorted by reverse of their subscription start date

veggiematts:
PR #501 provides a fix for this. However, the sort order is subscription start date desc, subscription end date desc.

sbouis:
Ok, thanks :)

PR #501, opened 2018-10-15
Improvements to the order dropdown list

veggiematts:
- Sort orders by subscription start date desc, subscription end date desc
 - Select the first order that is currently active (start date <= today => end date)
 - Show the order's organization after the dates

Before:
orders_before

After:
orders_after

### DIFF
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -108,7 +108,7 @@ class Resource extends DatabaseObject {
     }
 
     public function getResourceAcquisitions() {
-        $query = "SELECT * from ResourceAcquisition WHERE resourceID = " . $this->resourceID;
+        $query = "SELECT * from ResourceAcquisition WHERE resourceID = " . $this->resourceID . " ORDER BY subscriptionStartDate DESC, subscriptionEndDate DESC";
 		$result = $this->db->processQuery($query, 'assoc');
         $objects = array();
 
--- a/resources/admin/classes/domain/ResourceAcquisition.php
+++ b/resources/admin/classes/domain/ResourceAcquisition.php
@@ -986,7 +986,14 @@ class ResourceAcquisition extends DatabaseObject {
 		return $objects;
 	}
 
-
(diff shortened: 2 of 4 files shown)


veggiematts:
Fix for #451

poacherone:
We have tested this PR and it works as intended. The sorting makes sense, and the fact that you "land" on the current order is a good improvement (you might be adding new orders for the coming year, but the one still active should be the one currently used)!

The only tiny issue we faced was that the dropdown box doesn't have a fixed length, so if it is a very long organization name it will extend to cover the full name which makes it look a bit strange. But width issues are found elsewhere in Coral, so perhaps this should not be covered by this PR but be a separate (bigger) issue?

veggiematts:
Thank you for the feedback!

> The only tiny issue we faced was that the dropdown box doesn't have a fixed length, so if it is a very long organization name it will extend to cover the full name which makes it look a bit strange

In fact, the dropdown has a max size of 1000px, which is roughly the available length in Coral interface. So having very long organization names will make it look perhaps a bit strange, but not broken :)

sbouis:
Hello,
why do we add the organization names here? Is it from another PR or is it from this one?
For me, it's not necessary, but perhaps other libraries need it.

sbouis:
Ok, just read the beginning of this PR : "Show the order's organization after the dates"
Is it asked by other libraries?

veggiematts:
Yes, it was. This dev was sponsored by one of our customers.

sbouis:
Ok, I just wonder.

veggiematts:
@jeffnm I would be in favor of adding this to 3.0.1, even if it is not a bugfix, but it is a small enhancement that improves display and usability. What do you think?

jeffnm:
@veggiematts I'm also in favor, I've added it to the project and tested it. It works as described. Is there any implication from #489 for this? If you don't see any issues there, then let's merge it.

veggiematts:
> Is there any implication from #489 for this

No, #489 can (and should) be addressed separately.

> If you don't see any issues there, then let's merge it.
Let's do this :)

Issue #470, opened 2018-08-22
Fatal Error exporting Resource search results to Excel


jeffnm:
This was first reported during troubleshooting of #468 and I have been able to reproduce it myself.

Steps to reproduce:
1. Open CORAL 3.0 Resources module
2. (Optional) On the Home page do a search.
3. On the Home page click the Excel Icon in the top right corner.  
4. The page will return a 500 error. 

The Apache/PHP logs will say 

> `Fatal error: Uncaught exception 'Exception' with message 'There was a problem with the database: Unknown column 'RPAY.priceTaxExcluded' in 'field list'' in /var/www/html/coral/resources/admin/classes/common/DBService.php:52 Stack trace: #0 /var/www/html/coral/resources/admin/classes/common/DBService.php(91): DBService->checkForError() #1 /var/www/html/coral/resources/admin/classes/domain/Resource.php(1058): DBService->processQuery('SELECT R.resour...', 'assoc') #2 /var/www/html/coral/resources/export.php(46): Resource->export(Array, 'TRIM(LEADING 'T...') #3 {main} thrown in /var/www/html/coral/resources/admin/classes/common/DBService.php on line 52`

veggiematts:
I can experience the same problem on one of my instances, but it's RPAY.fundID that is missing.

This one can easily be explained: install.sql is wrong: table ResourcePayment should contain:

`  fundID int(11) default NULL,`

and not

`  fundName varchar(45) default NULL,`

@jeffnm could you paste a desc of your ResourcePayment table please ?

jeffnm:
It looks like the fundID change is in https://github.com/coral-erm/coral/blob/master/resources/install/protected/2.0.0/001-000.sql (used to be update_2.0.0.sql before the sequential SQL update model in 3.0.0). 

Here's one of our ResourcePayment tables. 

```
+-----------------------+------------------+------+-----+---------+----------------+
| Field                 | Type             | Null | Key | Default | Extra          |
+-----------------------+------------------+------+-----+---------+----------------+
| resourcePaymentID     | int(11)          | NO   | PRI | NULL    | auto_increment |
| resourceAcquisitionID | int(11) unsigned | NO   | MUL | NULL    |                |
| fundID                | int(10)          | YES  | MUL | NULL    |                |
| selectorLoginID       | varchar(45)      | YES  |     | NULL    |                |
| paymentAmount         | int(10) unsigned | YES  |     | NULL    |                |
[...]

PR #503, opened 2018-10-18
Issue 470 export resources bug

jeffnm:
This should fix the #470 issue for CORAL installations that were upgraded from 1.x versions in the past. 

During the 3.0.1 upgrade an SQL procedure will run to check if the columns in question are present or missing. If they are missing the columns will be added. 

The most significant change with this PR is that the installer's `db_tools.php` will now use `mysqli_multi_query` to run the entire SQL file, rather than attempting to split it into individual SQL statements with PHP's explode function. 

This PR also includes the first skeleton install helpers for 3.0.1

### Part 1
1. Start with a CORAL database and installation that is missing the `priceTaxExcluded`, `priceTaxIncluded`, and `taxRate` columns on the ResourcePayment table. This should be the state of any CORAL installation that was upgraded from 1.x to 2.0 or 3.0. Fresh installations to 2.0 or 3.0 will have these columns.
2. Attempt to download the Excel file export per #470. It will fail with a server error (blank screen).
3. Apply this pull request and run through the CORAL upgrade. 
4. Attempt step 2 again. The Excel file should download with the Resources search results you were displaying.

### Part 2
Complete a fresh installation of CORAL using this pull request to verify that there are no problems with the new `db_tools.php`


jeffnm:
I'm labeling this release blocking for 3.0.1. This PR needs to be included for 3.0.1 since it contains installer updates as well. 

If someone else could look at it and verify one or both of the install/update cases, I'd appreciate it.

LibClaire:
Hi, I tried to test this but was unsuccessful. Maybe it won't work the way I tested it which was by deleting the columns from my 3.0.0 database. 
After the columns were deleted I could reproduce the error with the export.
I applied the patch and the installer appeared asking me to grant and revoke permissions for 2 configuration.ini files and was then done. However the columns have not been added to the database and I still cannot export resources. I get the same php error as before.
LIke I say, perhaps I cannot test it this way. I checked and all commits were successfully applied. My Coral version is now '3.0.1' and the footer says development.

poacherone:
I tried part 1 of this PR over the weekend. I cheated a little an simply deleted the columns from my table in my 3.0.0 install, applied the PR and ran the Coral upgrade. It looks like it runs through the upgrade (I'm asked for db credentials etc.) but the columns aren't added to the database. Is this because I didn't do an installation of 1.x and then upgraded? Does the installer check versions? 

Or is it an issue with the SQL query? (I also tried running it directly, but that didn't work either.)

poacherone:
So basically, I experienced the same as @LibClaire reported above. ^^

Also have version and footer changed.

jeffnm:
Thanks @LibClaire and @poacherone let me look into it. It's supposed to check if those columns are present and, if they aren't, add them, which did work on my test server, but it might be something specific to my version or my database. 

It sounds like the installer processed, but the SQL file didn't get applied for some reason.

jeffnm:
It looks like it's a collation mismatch error. I think I can fix that, but this is revealing that the error handling for the mysqli_multi_query is not working properly.

jeffnm:
I pushed a commit that has better error handling for the mysql error messages so we should actually catch the error message, as well as making the parameter explicitly utf8 which solved the mismatch for my testing.

jeffnm:
@LibClaire and @poacherone would you be able to test this again? I think making the collation explicitly utf8 should have fixed it, but I definitely would like some more tests since it didn't work for you two previously.

poacherone:
Tested again and the installer now runs properly, and the database is updated with the missing columns. The export however, still doesn't work (blank page in Firefox, http 500 error in Chrome).

LibClaire:
Seems to be working. First attempt - worked straight away so I wanted to try a second test to be sure. 

I deleted the columns from ResourcePayment - I reapplied the patch but it didn't work, it seems I had to also change the version in common/configuration.ini back to 3.0.0 as it was still showing as development. 
Once I did that the installer ran again as expected, the columns were added to the table, version is set to 3.0.1 and I can download the file.

Looks good to me

poacherone:
After @LibClaire's comment, I went back and looked more closely at the logs. The export failed because there was missing columns in the Resource-table (currentStartDate and currentEndDate). The were missing because of testing [Pr 510](https://github.com/coral-erm/coral/pull/510), which previously dropped those tables if I recall correctly. Since this has changed, and the columns aren't dropped, it shouldn't affect this PR.

So, works as intended!

jeffnm:
Thank you both, @LibClaire and @poacherone. It's much appreciated!

I'm concerned that the export requires the Resource table currentStartDate and currentEndDate columns, when those are considered to be deprecated now, but I guess that's a different issue. 

@veggiematts can you merge this based on those two tests? I don't want to merge my own PR, especially when it is a little more involved like this one.

templates/footer.php
```diff
@@ -24,7 +24,7 @@
 <div class="push">&nbsp;</div>
 </div>
 
-<div class="footer"><?php echo _("Copyright");?> &copy; 2018. <?php echo _("CORAL version");?> 3.0.0<br/>
+<div class="footer"><?php echo _("Copyright");?> &copy; 2018. <?php echo _("CORAL version");?> Development Branch<br/>
```
veggiematts: Is this change intended?

PR #505, opened 2018-10-23
Fix licensing calendar bad query failure

### LINKED ISSUE #461: CORAL 3.0 bad query failure for Calendar in the Licensing Module
CORAL user reported the calendar in the Licensing module does not work. After clicking on the Calendar tab, there is a message "Bad Query Failure" appears. I was able to reproduce the error on CORAL demo site and on another CORAL instance.

jeffnm:
This fixes  #461


jeffnm:
I've opened an issue on the documentation repository about the lack of calendar instructions. https://github.com/coral-erm/documentation/issues/51

Issue #478, opened 2018-08-31
Subscription Start and End dates cannot be unset in Resource -> Edit Subscription Info


sekjal:
In earlier versions of the code, unsetting the Start and End dates to a subscription would fail because the value was passed literally as NULL, and DatabaseObject didn't properly interpret that.

Then, in commit coral@7b2c86d, NULL is replaced by date("Y-m-d").  While this will now properly save, it sets the value to today's date, which doesn't make sense as a response to unsetting the date value entirely.

If we pass an empty string as the value for these attributes when they're unset in the form, the expected behavior of nullifying the values occurs.

sekjal:
And now testing in CORAL 3.0, it's even worse.  subscriptionStartDate and subscriptionEndDate cannot be NULL, even for Resources that aren't subscriptions and don't have start and end dates.

poacherone:
In Coral 3.0.0 subscriptionStartDate and subscriptionEndDate seem to get now() as default. And as described by @sekjal, this makes this issue even worse. You get a strange value (why would EndDate be now?) that you cannot unset.

Preferably, EndDate shouldn't be set by default at all. But most importantly: there needs to be a way to unset it!

A use case: We could have a resource that is a trial, and thus have an EndDate. The same resouce might then be change to paid if we decde to purchase it. The paid resource could be an ongoing, that shouldn't have an EndDate. We would then like to be able to un-select the EndDate value. I'm sure there are similar use-cases where youwould want to uncheck the StartDate as well.

PR #509, opened 2018-10-26
Issue 478: Allow subscription start and end date to be null

veggiematts:
Fix for #478 

Test plan:
  - apply resources/install/protected/3.0.1/001-478.sql
  - Check that you can unset an order subscription start date
  - Check that you can unset an order subscription end date
  - Check that you can unset both
  - Check that the display is correct in the order dropdown list
  - Check that the active order is correctly selected

issue_478

When both dates are unset, the order id is displayed, so the user can differentiate the orders.


keelyward:
This is a feature we are very excited to see in 3.0.1.

PR #510, opened 2018-10-29
Issue 489 fix subscription date fields

### LINKED ISSUE #489: Subscription dates not set correctly for orders during upgrade

jeffnm:
This should fix #489 

It includes SQL to copy the `currentStartDate` and `currentEndDate` fields from the Resource table to  the ResourceAcquisition table's `subscriptionStartDate` and `subscriptionEndDate`.  

There are also changes to code that utilizes those fields in the Resource table and class.

## Testing the SQL
Since this pull request does not include the 3.0.1 upgrade code, the SQL will not be run automatically. To test it, you will need a database that has the dates in Resource `currentStartDate` and `currentEndDate` and the automatically created dates in ResourceAcquisition `subscriptionStartDate` and `subscriptionEndDate`. When you apply the SQL file (`3.0.1/002-489.sql`) it will copy the columns and then drop the old ones. 

## Testing the other changes
Make sure you can see all the Order dates you expect in Resources and in Licensing's Calendar.


poacherone:
Tested and works as intended, as far as I can see (dates updated and displayed correctly). Note that I haven't got anything in my calendar in Licensing, but I am not sure how to set that up.

veggiematts:
Quick thoughts:
  - resource dates are still used in summary.php (resource -> Helpful Links -> print view). We should fix this.
  - shouldn't we also modify `install/protected/3.0.0/002-350.sql` so people that haven't upgraded to 3.0.0 yet will directly have the correct dates?

jeffnm:
I pushed changes to keep the old columns, revert the API to access those old columns, and fix the summary.php (print view) display.

The question of whether to modify the `install/protected/3.0.0/002-350.sql` file or not is interesting. My initial thought was no, the 3.0.1 patch will be immediately applied for anyone who is installing 3.0.0 for the first time. However, I think copying everything at the time of install/upgrade will be better. I'll make those changes.

veggiematts:
Tech review ok for me.

PR #511, opened 2018-11-08
Fix import page translation

veggiematts:
Check that the $verb variable and the sentences it is used in can correctly be extracted by gettext.

See https://github.com/coral-erm/coral/pull/507#pullrequestreview-172217651

### DIFF
--- a/resources/import.php
+++ b/resources/import.php
@@ -903,9 +903,9 @@
 			}
 			print "<h2>"._("Results")."</h2>";
 			$verb = isset($proceed) ? _("have been") : _("will be");
-			print "<p>" . ($row - 1) . _(" rows $verb processed. ").$inserted._(" rows $verb inserted.")."</p>";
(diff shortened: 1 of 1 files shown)


jeffnm:
Is it ok to split parts of a phrase up into separate translation pieces like that? Won't that make it harder for a natural translation since they have to follow the same order?

veggiematts:
> Won't that make it harder for a natural translation since they have to follow the same order?
Indeed. But apart from duplicating all the sentences, I didn't see a way to allow proper translation. Do you?

jeffnm:
> > Won't that make it harder for a natural translation since they have to follow the same order?
> 
> Indeed. But apart from duplicating all the sentences, I didn't see a way to allow proper translation. Do you?

Fair enough. A working but awkward translation is better than a broken one. 

I can merge it.

PR #514, opened 2018-11-28
use ResourceAcquisition instead of Resource for subscription dates in…

jeffnm:
… licensing calendar

An additional fix related to #461 

The calendar now checks the ResourceAquisition table rather than the Resource table when looking for subscription dates.


veggiematts:
Merging #510 caused conflicts, can you resolve them, @jeffnm ?

jeffnm:
This PR is no longer necessary. #510 fixed these issues.

PR #516, opened 2018-12-05
Add library number to orders and dashboards

veggiematts:
Test plan:
      - Apply the database update
      - Check that you can add/edit/remove a library number in orders
      - Check that you can group by library number in the dashboards

librarynumber
librarynumberstats


sbouis:
Installed and tested on our Coral insatllation. It works as expected.

sbouis:
This enhancement is really usefull for us to allow us to add specific information that we have to deal with in France.

LibClaire:
Tested this following the plan and works as expected.

PR #518, opened 2018-12-17
Allow to specify date format in common/configuration.ini

veggiematts:
Currently, Coral's date format is hardcoded in common/common_directory.php
This patch allows to specify the date format in common/configuration.ini, with a fallback to the default date format if nothing is specified.

```
[settings]
date_format = "%m/%d/%Y"
```

### DIFF
--- a/common/common_directory.php
+++ b/common/common_directory.php
@@ -76,8 +76,14 @@ function format_date($mysqlDate) {
 	//make sure digit years matches for both directory.php and common.js
 
 	//SUGGESTED: "m/d/Y" or "d-m-Y"
-
-	return date("m/d/Y", strtotime($mysqlDate));
+    $config = new Configuration();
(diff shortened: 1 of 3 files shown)


poacherone:
Tested, and this PR works the same as changing the code in common/common_directory.php. It mostly works as intended on most pages where dates are displayed, but I have opened two separate issues: [Issue 520](https://github.com/coral-erm/coral/issues/520) and [Issue 521](https://github.com/coral-erm/coral/issues/521).

Since the behavior is the same in changing the code directly (locally) or applying this PR, I would argue that this works as intended and that new PR:s should be created for the new issues.

veggiematts:
In order to fix Issue #521, the date format is now like: "%m/%d/%Y"

jeffnm:
Looks good to me.

Issue #520, opened 2018-12-18
Datepicker not providing correct date format


poacherone:
When formating how dates are displayed (either changing common/common_directory.php or applying [PR 518](https://github.com/coral-erm/coral/pull/518)) the datepicker still uses the default format when choosing a date from the calendar. It saves ok, but it looks odd that the chosen date format ins't respected by the datepicker/calendar. This behavior is the same wherever the datepicker is used: orders, cost history, searching resources for date created, the effective date for a document in the License module etc.

veggiematts:
Before a patch is made, you can fix the datepicker's behavior by changing the `Date.format` value in each module's `js/common.js`:

```
//suggested: mm/dd/yyyy OR dd-mm-yyyy
Date.format = 'dd/mm/yyyy';

```

PR #523, opened 2018-12-28
Issue 520: Allow to modify Date.format in the main common.js file

veggiematts:
Improvement/fix for #520 : Date.format (used by datepickers) is now configurable in a single file (`js/common.js` ) (instead of having to configure it in every module)

### DIFF
--- a/licensing/js/common.js
+++ b/licensing/js/common.js
@@ -30,14 +30,6 @@
 })(jQuery)
 
 
-
-//Required for date picker
-Date.firstDayOfWeek = 0;
-
-//suggested: mm/dd/yyyy OR dd-mm-yyyy
-Date.format = 'mm/dd/yyyy';
-
-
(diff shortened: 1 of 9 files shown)


veggiematts:
Test plan:

1) Edit `Date.format` in `js/common.js`. It must be set accordingly to `date_format` in `common/configuration.ini`, otherwise dates won't be stored properly.

2) Check that the date format provided by the Datepickers in all the modules are correct.

3) Check that dates are stored properly.

poacherone:
Works as intended! Very nice!

The only (tiny) related issue I found is that in when editing Cost history there is a date format already filled in the empty row (in light gray), I suppose as a guide on how to write the date format. This date format isn't affected by the changes in this PR, nor [PR 518](https://github.com/coral-erm/coral/pull/518)

Issue #527, opened 2019-01-11
Search on notes in Resources module


sbouis:
Search on notes in Resources module is weird.
I’ve tried several searches on the demo website :
1.	I’ve selected “access details” in the field “note type” in the search column : I’ve found 23 records
2.	I’ve had “useful” in the field “Note (contains)” and click on the “go” button : I’ve found 12 records but when I’ve checked in the resources, a lot of them didn’t have the word “useful” in the notes.
3.	I’ve searched only for “useful” in the field “Note (contains)” and click on the “go” button : I’ve found one record.

I really don’t understand the second result…

The behavior I would expect is :
-	If we use only “Note Type” : find all the resources that have a note of that type
-	If we use only “Note contains” : find all the resources that have this term in all type of notes
-	If we use the field “Note type” and “Note contains” : find all the resources that have this term only in the type of notes selected

PR #528, opened 2019-01-14
Issue 527: Fix resource search on note type and note text.

veggiematts:
Fix for Issue #527 
Too many results were returned when searching on both note type and note text.

Test plan:
1) Apply the patch
2) Search resources using note type and note text
3) Check that the results are resources having a note of the searched type containing the searched text (and not a mix of both).


veggiematts:
Thank you Claire.

PR #536, opened 2019-02-13
issue #445, licensing: use ANY_VALUE() around shortName column

t4k:
see #445 

The underlying error from `PHP Warning:  mysqli_fetch_assoc() expects parameter
1 to be mysqli_result, boolean given in
…/licensing/admin/classes/domain/License.php on line 731, referer:
http://localhost:8080/coral/organizations/index.php`

is

`Expression #1 of SELECT list is not in GROUP BY clause and contains
nonaggregated column 'coral_licensing.L.shortName' which is not functionally
dependent on columns in GROUP BY clause; this is incompatible with
sql_mode=only_full_group_by`

Running the query directly in MySQL 5.7 will produce the error. Using the
ANY_VALUE() function suppresses the test for nondeterminism.

https://dev.mysql.com/doc/refman/5.7/en/miscellaneous-functions.html#function_any-value

PR #537, opened 2019-02-13
management: use ANY_VALUE() around shortName column

t4k:
see #445

Using the ANY_VALUE() function suppresses the test for nondeterminism.

https://dev.mysql.com/doc/refman/5.7/en/miscellaneous-functions.html#function_any-value

```
PHP Fatal error:  Uncaught exception 'Exception' with message 'There was a
problem with the database: Expression #1 of SELECT list is not in GROUP BY
clause and contains nonaggregated column 'coral_management.L.shortName' which is
not functionally dependent on columns in GROUP BY clause; this is incompatible
with sql_mode=only_full_group_by' in …/management/admin/classes/common/DBService.php:55
Stack trace:
#0 …/management/admin/classes/common/DBService.php(82): DBService->checkForError()
#1 …/management/admin/classes/domain/License.php(804): DBService->query('\\n\\t\\t\\tSELECT\\n\\t\\t\\t\\t...')
#2 …/management/admin/classes/common/Object.php(55): License->getAlphabeticalList()
#3 …/management/index.php(229): Object->__get('getAlphabetical...')
#4 {main}
  thrown in …/management/admin/classes/common/DBService.php on line 55,
  referer: http://localhost:8080/coral/licensing/
```

### DIFF
--- a/management/admin/classes/domain/License.php
+++ b/management/admin/classes/domain/License.php
@@ -797,7 +797,7 @@ class License extends DatabaseObject {
 		$alphArray = array();
 		$result = $this->db->query("
 			SELECT
-				DISTINCT UPPER(SUBSTR(TRIM(LEADING 'The ' FROM shortName),1,1)) letter,
+				DISTINCT UPPER(SUBSTR(TRIM(LEADING 'The ' FROM ANY_VALUE(shortName)),1,1)) letter,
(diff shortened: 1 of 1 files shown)


t4k:
As noted in #536 the use of ANY_VALUE() is incompatible with MariaDB. This needs work.

jeffnm:
@t4k can this be tested or does it still need more work?

t4k:
@jeffnm I believe the last commit should have it working. It can be tested.

Issue #543, opened 2019-02-27
Licensing, Expressions: column 'lastUpdateDate' cannot be null


t4k:
I'm posting the contents of my email to the coral-users list because I think it's a legitimate issue after investigating further. <https://lists.coral-erm.org/pipermail/coral-user/2019-February/000506.html>

We are getting an error when trying to save an Expression on a license.

> There was a problem with the database: Column 'lastUpdateDate' cannot be null

From what I can tell in the code, this makes sense…

https://github.com/coral-erm/coral/blob/97526009bd1035b5d293741195c90265726cacab/licensing/install/protected/install.sql#L51 declares that the column in the db cannot be NULL.

https://github.com/coral-erm/coral/blob/97526009bd1035b5d293741195c90265726cacab/licensing/admin/classes/common/DatabaseObject.php#L223-L224 sets the value of the attribute to NULL if it is sent with nothing.

https://github.com/coral-erm/coral/blob/97526009bd1035b5d293741195c90265726cacab/licensing/ajax_processing.php#L259-L303 does not set lastUpdateDate anywhere.

I'm puzzled why I cannot reproduce the error on any test instances, only on our production instance.

The production instance is using:
PHP Version 7.0.33-0ubuntu0.16.04.1
MySQL Server version: 5.6.40-log Source distribution (AWS RDS)
Apache/2.4.18 (Ubuntu)

t4k:
I have discovered what the problem is for our setup.

AWS RDS MySQL sets the `explicit_defaults_for_timestamp` option to TRUE in its default parameter group for MySQL 5.6. It can only be changed by someone going in and adjusting a custom parameter for that setting to FALSE and restarting the database service.

Although this is an edge case, I believe the CORAL code can be improved to overcome it. It is very possible that someone may come along and attempt to use AWS RDS with MySQL 5.6 with default settings and hit the same problem as I did.

PR #544, opened 2019-02-27
Issue #543: set $expression->lastUpdateDate

t4k:
The real changes besides white space include:

- set `$expression->productionUseInd` to a string in the `else` clause
- avoid rewriting `$expression->productionUseInd` immediately after the `else` clause (if it is an update and the value is already set, it would have been always overwritten)
- set `$expression->lastUpdateDate = date('Y-m-d H:i:s');` so that we don't get the cannot be null error

Issue #543 

# TEST PLAN

Environment (I believe the problem lies in the MySQL configuration):

- PHP Version 7.0.33-0ubuntu0.16.04.1
- MySQL Server version: 5.6.40-log Source distribution (AWS RDS)
  - `explicit_defaults_for_timestamp` set to `TRUE`
- Apache/2.4.18 (Ubuntu)

## Confirmation of Problem
Without the patch, when trying to add an Expression to a License in the Licensing module, the error *There was a problem with the database: Column 'lastUpdateDate' cannot be null* is shown and the Expression cannot be saved.

## Resolution
With the patch, the `lastUpdateDate` value is explicitly set so that the `DatabaseObject` code is not finding no value and trying to set the field to `NULL`. The Expression should save without any problems.

### DIFF
--- a/licensing/ajax_processing.php
+++ b/licensing/ajax_processing.php
@@ -255,52 +255,51 @@ switch ($_GET['action']) {
 		break;
 
 
-	//add/update expression
+    // add/update expression
     case 'submitExpression':
 
-    	//if expressionID is sent then this is an update
-    	if ((isset($_POST['expressionID'])) && ($_POST['expressionID'] != '')){
-    		$expressionID = $_POST['expressionID'];
- 			$expression = new Expression(new NamedArguments(array('primaryKey' => $expressionID)));
-    	}else{
- 			$expression = new Expression();
- 			//default production use (terms tool indicator) to off if this is an add, otherwise we leave it alone
-			$expression->productionUseInd	= 0;
-			$expression->expressionID = '';
-		}
+      // if expressionID is sent then this is an update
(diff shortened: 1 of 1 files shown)


jeffnm:
Hey @t4k, could you suggest a testing plan for this PR?

t4k:
I found out why this is so difficult to reproduce. It has to do with the `explicit_defaults_for_timestamp` that is set to `TRUE` by default on AWS RDS MySQL 5.6.

When this value is set we run into a problem because the code
https://github.com/coral-erm/coral/blob/97526009bd1035b5d293741195c90265726cacab/licensing/admin/classes/common/DatabaseObject.php#L223-L224
is trying to save a `NULL` value for the field. Now this would be normally overcome by the schema setting of
https://github.com/coral-erm/coral/blob/97526009bd1035b5d293741195c90265726cacab/licensing/install/protected/install.sql#L51
but it seems that the MySQL configuration overrides that schema setting.

streatim:

PR #546, opened 2019-03-06
Issue 387 remplacing links by icons

Malmousque:
This PR replaces links, particulary in tabs, by icons into the 5 modules, some icons was already in use in some place, the idea is to improve and make the UX coherent.

coralicons3

5 news CSS classes have been added : .removeIcon  .editIcon  .updateIcon .addIcon .addIconTab.
All the icons was already in use. Most of them was only in the Resources modules images folder, I add them in the others images folders.

To update
coraladd

To add
coraladding
 
To remove
coralremove

To edit
coraledit

Is the update and edit function have a real separate function ?


licensing/ajax_htmldata.php
```diff
@@ -203,7 +203,7 @@
 				echo "<td><div id='attachment_short_" . $attachment->attachmentID . "'>" . substr($attachmentText, 0,200);
 
 				if (strlen($attachmentText) > 200){
-					echo "...&nbsp;&nbsp;<a href='javascript:showFullAttachmentText(\"" . $attachment->attachmentID . "\");'>"._("more...")."</a>";
+					echo "...&nbsp;&nbsp;<a href='javascript:showFullAtt		break;achmentText(\"" . $attachment->attachmentID . "\");'>"._("more...")."</a>";
```
veggiematts: Typo ?

licensing/js/admin.js
```diff
@@ -354,7 +354,7 @@ function submitQualifier(){
 
 
 function showAdd(tableName){
-       $('#span_new' + tableName).html("<input type='text' name='new" + tableName + "' id='new" + tableName + "' class='adminAddInput' />  <a href='javascript:addData(\"" + tableName + "\");'>"+_("add")+"</a>");
+       $('#span_new' + tableName).html("<input type='text' name='new" + tableName + "' id='new" + tableName + "' class='adminAddInput' />  <a href='javascript:addData(\"" + tableName + "\");'><img id='addCurrency' src='images/plus.gif' title='add' /></a>");
```
veggiematts: Translation

resources/ajax_htmldata/getAccessDetails.php
```diff
@@ -158,19 +158,23 @@
...
 				<tr>
-				<th><?php echo _("Additional Notes");?></th>
+				 <th><?php echo _("Additional Notes");?></th>
 				<th>
+
 				<?php if ($user->canEdit()){?>
-					<a href='ajax_forms.php?action=getNoteForm&height=233&width=410&tab=Access&entityID=<?php echo $resourceAcquisitionID; ?>&resourceNoteID=&modal=true' class='thickbox'><?php echo _("add new note");?></a>
+				<a href='ajax_forms.php?action=getNoteForm&height=233&width=410&tab=Access&entityID=<?php echo $resourceAcquisitionID; ?>&resourceNoteID=&modal=true' class='thickbox'><?php echo "<div class= 'addIconTab'><img id='Add' class='addIcon' src='images/plus.gif' title= '"._("Add")."' /></div>";?></a>
+				<?php }
+				if ($user->canEdit()){?>
```
veggiematts: This seems misplaced.

resources/ajax_htmldata/getAdminAlertDisplay.php
```diff
@@ -35,10 +38,15 @@
...
 		}
 
-		echo "<a href='ajax_forms.php?action=getAdminAlertEmailForm&alertEmailAddressID=&height=128&width=260&modal=true' class='thickbox'>"._("add email address")."</a>";
-		echo "<br /><br /><br />";
 
 
+		?>
+		<div class="adminHeader">
+			<div></div>
+			<div class="addElement" style= "padding-bottom: 8px;"><?php echo "<a href='ajax_forms.php?action=getAdminAlertDaysForm&alertDaysInAdvanceID=&height=128&width=260&modal=true' class='thickbox'><img id='addAlertDay' src='images/plus.gif' title='add day'/></a>";?></div>
```
veggiematts: Translation

resources/ajax_htmldata/getAdminUserDisplay.php
```diff
@@ -16,7 +16,10 @@
 
 		if (count($instanceArray) > 0){
 			?>
+			<div class="adminHeader">
 			<div class="adminRightHeader"><?php echo _("Users");?></div>
+			<div class="addElement" style="margin-right: 4px"><a href='ajax_forms.php?action=getAdminUserUpdateForm&loginID=&height=275&width=315&modal=true' class='thickbox' id='addUser'><img id='addUserGroup' src='images/plus.gif' title='add User' /></a></div>
```
veggiematts: Translation

resources/ajax_htmldata/getProductDetails.php
```diff
@@ -434,17 +434,21 @@ class='removeResourceSubjectRelationship'
 				<tr>
 				<th><?php echo _("Additional Notes");?></th>
 				<th>
+
 				<?php if ($user->canEdit()){ ?>
-					<a href='ajax_forms.php?action=getNoteForm&height=233&width=410&tab=Product&entityID=<?php echo $resourceID; ?>&resourceNoteID=&modal=true' class='thickbox'><?php echo _("add new note");?></a>
+					<a href='ajax_forms.php?action=getNoteForm&height=233&width=410&tab=Product&entityID=<?php echo $resourceID; ?>&resourceNoteID=&modal=true' class='thickbox'><?php echo "<div class='addIconTab' ><img id='Add' src='images/plus.gif' title= '"._("Add")."' /></div>";?></a></div>
+					<?php if ($user->canEdit()){ ?>
```
veggiematts: This seems misplaced

LibClaire:
Hi, 
I've just had a first look at this. Patch applies without a problem and the icons are working as links.

I have a couple of comments: 
- Resources module: it would be nice if the "add" icon appeared next to the headings as it does in the other modules. I think it reads better that way and it is clearer what the icon is for. 

- Resources module: Under admin - alert settings there are two "add" icons, one for "add email" and one for "add day", it might be a good idea to have some text here to help identify them. (I have seen the title hover text, which is good, but I don't think it is enough when there is more than one option)

- Licensing module: we have two different options, update and edit. My personal opinion is that we don't need to have two different icons/definitions here as they do the same thing? I came across this recently while translating and I think "edit" would be enough for all cases. Happy to have further feedback on this, perhaps there was a use case for it.

Otherwise I think this is good! I like the icons. :)

Malmousque:
Hi,
Sorry for not answering faster. The "add" icon now appears next to the headings in the ressource module, and I have added text to identify "add" icon in alert settings.
It would be maybe better to keep the question of "update/edit" for a next issue.
Thank you for your recommendations. :)

resources/ajax_htmldata/getAdminAlertDisplay.php
```diff
@@ -42,8 +42,8 @@
 
 		?>
 		<div class="adminHeader">
-			<div></div>
-			<div class="addElement" style= "padding-bottom: 8px;"><?php echo "<a href='ajax_forms.php?action=getAdminAlertDaysForm&alertDaysInAdvanceID=&height=128&width=260&modal=true' class='thickbox'><img id='addAlertDay' src='images/plus.gif' title='"._("add day")."'/></a>";?></div>
+			</div>
+			<div class="addElement" style= "padding-bottom: 8px;"><?php echo "<div><span class= 'addIconAlert'>Add a day: &nbsp;</span><a href='ajax_forms.php?action=getAdminAlertDaysForm&alertDaysInAdvanceID=&height=128&width=260&modal=true' class='thickbox'><img id='addAlertDay' src='images/plus.gif' title='"._("add day")."'/></a></div>";?></div>
```
veggiematts: Translation

resources/ajax_htmldata/getAdminAlertDisplay.php
```diff
@@ -7,9 +7,9 @@
 		$emailAddressArray = $alertEmailAddress->allAsArray();
 		$daysInAdvanceArray = $alertDaysInAdvance->allAsArray();
 		?>
-		<div class="adminHeader">
-		<?php echo "<div class='adminRightHeader'>"._("Alert Settings")."</div>";?>
-		<div class="addElement"><?php echo "<a href='ajax_forms.php?action=getAdminAlertEmailForm&alertEmailAddressID=&height=128&width=260&modal=true' class='thickbox'><img id='addAlertEmail' src='images/plus.gif' title='"._("add Email")."' /></a>";?></div>
+		<div class="adminHeaderAlert">
+		<?php echo "<div class='adminRightHeader'>"._("Alert Settings")."</div><br>";?>
+		<div class="addElement"><?php echo "<div><span class= 'addIconAlert'>Add an email: &nbsp;</span><a href='ajax_forms.php?action=getAdminAlertEmailForm&alertEmailAddressID=&height=128&width=260&modal=true' class='thickbox'><img id='addAlertEmail' src='images/plus.gif' title='"._("add Email")."' /></a></div>";?></div></br>
```
veggiematts: Translation

Malmousque:
"Add an email" and "Add a day" in the alert settings / ressources module can, now, be translated.

LibClaire:
Hi Malmousque,
the changes look great! 
I only have one other comment/question
Funds: here we have the 'down arrow' for an import. I am wondering if it is immediately clear what the icon means? To me an 'up arrow' would make more sense for an import and a 'down arrow' for download,  but I know we have that in the other modules as 'update'... maybe we can address it in a new pull request but I wanted to flag it as possibly confusing.

Malmousque:
Hi LibClaire,

I'm agree with you. For now, I have done with the icons provided with coral, but we could use others. I will talk about that with matts.

veggiematts:
I think we can safely merge 'update' and 'edit' buttons to only have 'edit'.
I think having both wasn't made on purpose, but comes from development. We have two types of tables for admin preferences: standard and custom ones. All the standard ones use 'edit', and some of the custom ones use 'update'. We should make this consistent and only keep 'edit'.

PaulPoulain:
++ for merging update & edit and thus be consistent !

LibClaire:
I would agree to just having 'edit'. I think it will be cleaner that way

LibClaire:
Hi,
everything is looking good. The import icon is better! 
One last comment. When going to edit something in a module that isn't resources, the heading of the pop up is still 'update'. Do we want to change the wording so that it is consistent?

Malmousque:
Hi LibClaire,

Yes, of course, I had to miss some "update".
Thanks

Malmousque:
Hi,
Sorry Claire, but I can't find the pop up you're talking about. All the 'Edit' icons seem to have an 'Edit' pop up. Could you tell me where it is ?

veggiematts:
Editing users in licensing for instance:
Capture d’écran_2019-03-29_13-52-42
It should be "Edit user"

Malmousque:
I modified the titles of the windows and harmonized the titles of the buttons (because for some, the title of the button and the window was in the same variable).
Have a nice WE !

LibClaire:
Hi Malmousque,
I think it looks good now. :)

veggiematts:
Merging this to devel. Thank you @LibClaire and @Malmousque for the good work!

PR #547, opened 2019-03-08
Fix licensing calendar

veggiematts:
The calendar page in the licensing module is currently broken: it never returns any data.
This patch intends to fix this.

Please note that this is a hotfix. It does return some data now, but I didn't check that said data is what is expected.

Please test this, and report any functional problem.

### DIFF
--- a/licensing/calendar.php
+++ b/licensing/calendar.php
@@ -138,16 +138,17 @@ $result = mysqli_query($link, $query) or die(_("Bad Query Failure: ".mysqli_erro
 					  `$resource_databaseName`.`AuthorizedSite`.`authorizedSiteID`
 					FROM
 					  `$resource_databaseName`.`Resource`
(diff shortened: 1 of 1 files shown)


poacherone:
This PR joins the tables correctly, giving results in the table. But the actual dates (how long it has expired) doesn't show, only showing 0. To fix this line 195 in licensing/calendar.php should be changed from “currentEndDate” to “subscriptionEndDate”.

veggiematts:
@poacherone can you confirm the fix is ok with your modification ?

sbouis:
I tried this one.
After changing settings in Licensing admin (Authorized Sites; Days After Subscription End; Days Before Subscription End; Resource Types), I can confirm that with this correction I can see in Calendar the Upcoming License Renewals that I coudln't see without.
It lists the resources with a Current Sub End set  in an  order that is beetween  Current Sub End - Days After Subscription End and  Current Sub End + Days Before Subscription End.

I only noticed an encoding problem with french accents.

veggiematts:
Thanks for the feedback, @sbouis 
I'm merging this, can you create a separate issue for the encoding problem ?

Issue #550, opened 2019-03-12
ILS Connection: Vendor name not retrieved from Koha


LibClaire:
The full vendor name is not pulled through from Koha when retrieving vendor information from the ILS.

For Example, if I have a vendor in Koha called 'Book Supplier' and add a new vendor in CORAL I can type 'Boo' and check if it already exists in Koha. If I then retrieve the information it pulls through the relevant fields except the name and then greys it out leaving me with 'Boo' as the vendor name instead of Book Supplier.

It would be good if the full name was also pulled through to CORAL.

veggiematts:
Obsoleted by #551

PR #551, opened 2019-03-12
Issue 550: ILS Connection: Retrieve vendor name from Koha

veggiematts:
Retrieve vendor name from Koha along with the other vendor fields.
See Issue #550


veggiematts:
FIXME: Check that the organization does not already exist in Coral upon name completion.

veggiematts:
Obsoleted by #577

Issue #418, opened 2018-04-26
Cost history numeric error


poacherone:
When adding a cost that is above a million, Coral will add two commas to the line (only in the interface, it's not added to the database). But trying to edit the cost it gives an error: price not numeric.

![coral_cost_numeric_error](https://user-images.githubusercontent.com/12153423/39333659-edddeca4-49ab-11e8-804d-ac4d615924e7.png)

This can be worked around by deleting the extra punctuation when editing (they will be added anyway), but is annoying nevertheless. Extra interpunction should preferably not be added to the edit cost page at all! Especially taken into consideration the different ways of writing larger numbers internationally, as with issue #311

PR #555, opened 2019-03-12
Fixes issue #418: Cost history numeric error

sekjal:
Replaces all commas in a proposed amount, not just 1.  Only comes up with very expensive items....


poacherone:
Tested, and works like a charm! I would suggest adding another line though: 
pAmount = pAmount.replace(/\s/g,'');

This will also fix any whitespace characters, in case they where added by mistake or (as in our case) by choice of currency display.

veggiematts:
"This will also fix any whitespace characters, in case they where added by mistake or (as in our case) by choice of currency display."

Can you elaborate on this ?

poacherone:
We display currency with a space for thousands. Like: "10 000", instead of 10,000. It is also easy to add a trailing whitespace by mistake (say "10,000 "). The second line will strip them.

veggiematts:
Agreed. @sekjal can you update your PR to include @poacherone proposal ?

veggiematts:
This should have been fixed with #576 .

Issue #556, opened 2019-03-12
Resources: Authorized Site duplication


t4k:
From @lnarizny: https://lists.coral-erm.org/pipermail/coral-user/2019-March/000507.html

> In attempting to add some authorized sites in the Access tab of a resource, I'm running into two problems: 1) duplication, and 2) the inability to remove an authorized site.

Steps to Reproduce:

>  1. Added some entries to the Authorized Sites list (including editing the default Main Site to have a new label) in Resources Admin
>  2. Went into my test resource record and found that one of the boxes (probably the former default Main Site) was already ticked
>  3. Ticked some more boxes and saved, only to find that a couple duplicate entries were produced
>  4. Unticked some boxes and saved, which did not remove any sites from the list and seemed to create a few more duplicates
>  5. Unticked all the boxes and saved, to no effect. I tried multiple different boxes different times

PR #557, opened 2019-03-12
556, make Authorized Sites compatible with multiple ordering

t4k:
Issue #556.

Allows Authorized Sites to be removed.


veggiematts:
Can confirm the erratic behavior without the patch.
Can confirm it works as expected with the patch.
Code review OK.
Merging this.

Issue #311, opened 2017-10-19
Numbers internationalization 


veggiematts:
Currently, in Coral, the only representation of numbers is the american one, but numbers representation depends on the country. For example:

The number 9988776.65 will be:
 - 9 988 776,65 in France
 - 9.988.776,65 in Germany
 - 9,988,776.65 in the United States

That's a problem, because:

1) It's counter-intuitive for the users to enter numbers, now that Coral is translated
2) exported data will be hard to process by other softwares that are aware of country-formatting specifics.

I made a proof of concept for the payment amounts in cost history, available here:
https://github.com/biblibre/Coral/tree/payment_amounts_i18n

It's not a pull request yet, because it is incomplete, and may have broken other things in Coral, and is only for the resources module right now, but I wanted your opinion about it.

It uses PHP NumberFormatter class: http://php.net/manual/en/class.numberformatter.php
It is a new dependency that would be needed for Coral, but it seems to be packaged in most distributions (package php-intl on ubuntu)

Please have a look at the code, it's a small patch, and NumberFormatter is really easy to use.

You can simply test it by changing the translation, and checking that the payment amounts are displayed according to the selected language.

If you want to modify payment amounts, it also works, but please note that the payment amount related javascript has been disabled for this POC, so the price will not be calculated from the tax rate, and the form will not check that the input is a number.

If you agree that this is the direction we should be headed to, I'll have a closer look on how to handle numbers i18n in javascript.

poacherone:
I haven't looked at the actual code at the repo that @veggiematts refers to, but in general this type of solution as absolutely needed. However, I am not too keen to see a solution where this is set by the selected language, simply because there is no 1 to 1 relationship between language and numbers representation. It would be best if this was configurable instead.

veggiematts:
"However, I am not too keen to see a solution where this is set by the selected language, simply because there is no 1 to 1 relationship between language and numbers representation. It would be best if this was configurable instead."

It could be done as language selection in web-apps: default language is based on what the browsers says it likes most, and then the user can choose another language.
In our case, the default format could be based on the user language, then changed if needed.

veggiematts:
So, to sum things up, we need to be able to use custom number formats in 3 places:

 - PHP code (using PHP NumberFormatter)
 - Javascript code (using ?)
 - SQL queries (like in resources exports and dashboards, for example) (using custom code?)

poacherone:
Yes that seems like a good summary.

[...]

PR #575, opened 2019-04-10
Allow to define SQL number format in common/configuration.ini

veggiematts:
Currently, all numbers displayed directly through SQL queries (exports and dashboards) are in the default format:

 - en_US if not set otherwise by the SQL server
 - 2 decimals

This patchs allows to define those preferences in common/configuration.ini, as such:

```
[database]
sql_locale = "fr_FR"
sql_decimals = 1
```

If `sql_locale` is omitted, it will be the SQL server default (en_US by default)
If `sql_decimals` is omitted, it will be defaulted to 2.

Please note that is PR is a partial solution to Issue #311 : Numbers internationalization.


poacherone:
Tested and works perfectly! Super job Matthias!

jeffnm:
If #576 is merged this won't be needed.

PR #576, opened 2019-04-11
Issue 311: Numbers internationalization

veggiematts:
Background & discussion, see Issue #311 

Currently, all numbers in Coral are in the default format:

 - en_US
 - 2 decimals

This patchs allows to define how numbers should be parsed and displayed in Coral.

First of all, install the `php-intl` package.

The configuration takes place in `common/configuration.ini`, as such:

```
[settings]
number_locale = "fr_FR"
number_decimals = 2
```

If number_locale is omitted, it will be defaulted to en_US
If number_decimals is omitted, it will be defaulted to 2

Please note that is PR includes and obsoletes:
 - PR #575 : Allow to define SQL number format in common/configuration.ini
 - PR #418 : Cost history numeric error

This PR currently affects every part of the resources module using
cost_to_integer and integer_to_cost function:
 - new resource, api, api_client, imports, summary and cost history.

Special sql processing has been made to exports and dashboards.
Special javascript processing has been made to cost history.

Test plan:
 1) Change the default values (as provided in the example above, for instance)
 2) Check that resource exports and dashboards show the correct number formatting
 3) Check that cost history fields correctly parses the numbers according to the
    locale when calculating priceTaxIncluded
 4) Check that cost history correctly displays the numbers according to the
    locale when calculating priceTaxIncluded
 5) Check that the numbers are correctly stored to and fetched from the database
    according to the locale in cost history
 6) Check that the import tool correctly parses the numbers according to the locale
 7) Check that it also works the default values or no values at all


poacherone:
I tried testing this PR but couldn't get it to work properly. In the dashboard and on exports, it works well, but the Resource Acquisitions tab shows a blank page whenever there are any actual costs (with no cost history, it displays fine). It looks like this:

Coral resource blank

I don't know if this is an issue with my test environment, or if anyone else can re-produce this?

veggiematts:
Any error in the logs ?

poacherone:
[php7:error] [pid 4155] [client 130.237.153.122:54600] PHP Fatal error:  Uncaught Error: Class 'NumberFormatter' not found in /var/www/html/coral/resources/directory.php:34\nStack trace:\n#0 /var/www/html/coral/resources/ajax_htmldata/getAcquisitionsDetails.php(144): integer_to_cost(NULL)\n#1 /var/www/html/coral/resources/ajax_htmldata.php(24): include('/var/www/html/c...')\n#2 {main}\n  thrown in /var/www/html/coral/resources/directory.php on line 34, referer: http://sysdev.sub.su.se/coral/resources/resource.php?resourceID=5808

veggiematts:
My bad, I forgot to include instructions to install the NumberFormatter dependency.
Installing the `php-intl` package should do the trick.

poacherone:
Thanks! Installing this did the trick. 

(One thing to note is that php-intl is currently defaulting at php 7.3, which might not work. I got it working only after downgrading to the installed php version, in my case 7.1)

veggiematts:
Rebased, and ready to be tested again :)

poacherone:
Impressive work! I followed the test plan and it mostly works as intended. That number_decimal_separator is needed in the config, while the actual display of the separator is taken from the number_locale doesn't make much sense from a user perspective, but might be required in the code?

One issue found is with the warning for non-numerical values when adding/editing a payment in cost history (in the Resource Acquisition tab). This doesn't seem to work at all with this PR. I've tested with +, / and a blank space, giving some unexpected results.

Coral_numbers_intl_cost_history

If placed where the thousand separator should be, it cuts off the value after the erraneous value. 23/000,00 will be saved as 23.00 (with number_decimals = 2). If placed anywhere else the number will be saved as 0. This is potentially **very** bad, since it means it will be very **easy to accidently delete costs**!

I would guess that this problem originates somewhere between number_locale and [PR 418](https://github.com/coral-erm/coral/issues/418). Perhaps with numbers formatting the need to remove commas and whitespaces is not an issue anymore?

To reprodue:
1. Add a number formatting, like number_locate= "fr_FR" and number_decimals=1
2. Go to cost history and add or edit a cost
3. Add extra whitespace characters, commas or characters/values not allowed (like + or /)
4. Notice that the cost/payment is cut off, or returned as 0

veggiematts:
> Impressive work! I followed the test plan and it mostly works as intended. That number_decimal_separator is needed in the config, while the actual display of the separator is taken from the number_locale doesn't make much sense from a user perspective, but might be required in the code?

Yes, that is because Javascript provides a function to format a number according a given locale, but doesn't provide a function to parse a number according a given locale. So I had to code one, and I couldn't find a way to retrieve the decimal separator from the number locale.

sbouis:
Thanks Matts. I've installed it on our test environnement but I have a problem in the resources module, in the acquisitions tab, there's just a blank screen : 
image

Perhaps, I have missed something...
I'm not sure to understand your note : 
Note: number_decimal_separator has to match the separator used in the number_locale locale.
Where could I check the separator used in the number_locale locale ?

poacherone:
Sonia, for the cost details to show properly you need to install the php-intl package.

sbouis:
Ok, thanks Andreas!

veggiematts:
> If placed where the thousand separator should be, it cuts off the value after the erraneous value. 23/000,00 will be saved as 23.00 (with number_decimals = 2). If placed anywhere else the number will be saved as 0. This is potentially very bad, since it means it will be very easy to accidently delete costs!

The input validation is now more strict: if the value contains anything else than digits, dots or commas, it won't be accepted.

veggiematts:
> Where could I check the separator used in the number_locale locale
Good news, I have found a way to get rid of  number_decimal_separator. I'm updating the test plan accordingly.

veggiematts:
@poacherone Hi, could you please have another look at this, to test the input validation and the removal of number_decimal_separator ?

poacherone:
Tested this again. The removal of the number_decimal_separator work very well! Makes it much easier to understand how to set-up.

The input validation might need some more work. As far as I can tell it only validates new additions (adding a new line and pressing "Add"), not the old entries. So all kinds of silly stuff can be done there to the old entries, including sums being deleted as I reported earlier.

I could also enter a whitespace in the new line, which I thought I couldn't do? (Only digits, dots and commas allowed?)

LibClaire:
I have tested this and it appears to be working. I tested with de_DE and en_GB. the input works correctly and rejects non numeric symbols in the numeric fields. Switching from one locale to another in the settings file also changed my data correctly in the GUI.

poacherone:
Also tested, and works as intended!

veggiematts:
Ok, just rebased this to development branch.
Since it has been tested ok by @poacherone and @LibClaire, I'm merging this.

poacherone:
Tested against the development branch, and works as intended. Good work @veggiematts!

PR #577, opened 2019-04-12
ILS Connection: use a dropdown list when searching an ILS vendor

veggiematts:
This patch allows the user to select an existing ILS vendor from a dropdown list:

vendors_dropdown

Test plan:
 - enable Coral <-> ILS connection
 - try to create a new Organization
 - check that a dropdown list containing ILS vendors appears
 - check that the ILS vendor informations are fetched
 - check that you can't create the organization if it already exists in Coral
 - check that you can create the organization if it doesn't already exist in Coral
 - disable Coral <-> ILS connection
 - check that the new Organization form behaves as it did before this PR.

If merged, this PR obsoletes PR #551 (issue #550)


LibClaire:
Tested and works as expected.

This is a really nice improvement for the ILS connection feature.

veggiematts:
Thanks for testing! Merging this.

Issue #581, opened 2019-04-25
Use config file for datepicker date format


veggiematts:
We shouldn't have to patch Coral to specify a date format for the datepickers (see PR #523) but rather be able to specify it in the main configuration file.

PR #582, opened 2019-04-25
Issue 581: Specify datepickers date format in common/configuration.ini

veggiematts:
Background: Issue #581 : We shouldn't have to patch Coral to specify a date format for the datepickers (see PR #523) but rather be able to specify it in the main configuration file.

Datepickers date format can be specified in common/configuration.ini, as such:

```
[settings]
datepicker_date_format = "dd/mm/yyyy";
```

If datepicker_date_format is omitted, it will be defaulted to 'mm/dd/yyyy'

### DIFF
--- a/common/common_directory.php
+++ b/common/common_directory.php
@@ -74,6 +74,17 @@ function return_date_format() {
     return $date_format;
 }
 
+function return_datepicker_date_format() {
+    $config = new Configuration();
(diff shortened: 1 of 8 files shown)


jeffnm:
This looks good to me. 

At some point we should add access for changing this setting in the Admin interface and/or in the Installer.

PR #591, opened 2019-07-31
Export Improvements

jcuenod:
## This PR provides two major improvements:

1. It speeds up the resources module by doing fewer joins and trying to only do joins that are necessary (which also helps with the export).
2. It supports exporting notes (as a json array)

## Requirements:

1. Run `resources/install/protected/next/001-16.sql ` against your development database. I'm not sure what needs to happen process-wise for this sql to be part of the next update cycle but that needs to happen...

## Things to test:

1. Go to `Admin` -> `Export Configuration` and try changing the settings navigating away and ensuring that they save.
2. With notes enabled, run an export and check that notes are included (probably the last column of the CSV). Check that the right notes appear.
3. Check that parent, child, and notes columns appear if and only if they are enabled in the export configuration options.
4. Under `Home`, try each of the search options. **Note:** It is worth checking that all these options still work correctly because this PR modifies joins and so while developing it, I definitely broke the search a couple of times.


resources/admin.php
```diff
@@ -30,90 +30,100 @@
...
+  if ($config->settings->organizationsModule == 'N'){
+    $adminMenuLinks[] = ["translatableTitle" => _("Organization Role"), "className" => "AdminLink", "id" => "OrganizationRole"];
+    $adminMenuLinks[] = ["translatableTitle" => _("Organizations"), "className" => "AdminLink", "id" => "Organization"];
+  }
+  // Sort the links according to their title (I do not believe that this is locale aware)
+  usort($adminMenuLinks, function ($item1, $item2) {
+    return strcmp($item1["translatableTitle"], $item2["translatableTitle"]);
+  });
+
+  // We want these links at the top of the list all the time
```
veggiematts: Are we sure we still want these links at the top of the list all the time ? I find it confusing.

resources/admin.php
```diff
@@ -30,90 +30,100 @@
...
+  if ($config->settings->enableAlerts == 'Y'){
+    $adminMenuLinks[] = ["translatableTitle" => _("Alert Settings"), "className" => "AlertAdminLink"];
+  }
+  if ($config->settings->organizationsModule == 'N'){
+    $adminMenuLinks[] = ["translatableTitle" => _("Organization Role"), "className" => "AdminLink", "id" => "OrganizationRole"];
+    $adminMenuLinks[] = ["translatableTitle" => _("Organizations"), "className" => "AdminLink", "id" => "Organization"];
+  }
+  // Sort the links according to their title (I do not believe that this is locale aware)
+  usort($adminMenuLinks, function ($item1, $item2) {
+    return strcmp($item1["translatableTitle"], $item2["translatableTitle"]);
```
veggiematts: I don't know if strcmp is locale aware, but this is a good start anyhow. Not a blocker for me. Can be improved later if needed.

resources/admin.php
jcuenod: Good question. I assumed that this was by design because everything else was alphabetical.

jcuenod:
Ja, sorry for the long PR. I could only see a way of cherry picking commits to create a PR and that wouldn't have made sense because some of the commits assume multiple parts of what's going on here. If you know of a way of creating PRs based on lines of code that have been changed in between two branches, I can work on putting together a couple of those...

veggiematts:
Here's what I've tested and seems OK:
 - SQL update
 - Export configuration page
 - Enabling/disabling export options
 - Notes export as JSON. Order-level notes are not exported, only resource-level notes.

Here's what I managed to break :)
 - Partial export when selecting a note type:
`Unknown column 'RNA.noteTypeID' in 'where clause' in /home/coral/www/resources/admin/classes/common/DBService.php:52
Stack trace:
#0 /home/coral/www/resources/admin/classes/common/DBService.php(91): DBService->checkForError()
#1 /home/coral/www/resources/admin/classes/domain/Resource.php(1241): DBService->processQuery('\nSELECT\n  R.res...', 'assoc')
#2 /home/coral/www/resources/export.php(47): Resource->export(Array, 'R.createDate DE...')`
 - All exports are broken after you click on "New search"
`2019/08/02 15:50:32 [error] 399#399: *625 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught Exception: There was a problem with the database: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '' at line 156 in /home/coral/www/resources/admin/classes/common/DBService.php:52
Stack trace:
#0 /home/coral/www/resources/admin/classes/common/DBService.php(91): DBService->checkForError()
#1 /home/coral/www/resources/admin/classes/domain/Resource.php(1241): DBService->processQuery('\nSELECT\n  R.res...', 'assoc')
#2 /home/coral/www/resources/export.php(47): Resource->export(Array, '')`
 - Partial export when filtering by note content:
`Unknown column 'RNA.noteText' in 'where clause' in /home/coral/www/resources/admin/classes/common/DBService.php:52
Stack trace:
#0 /home/coral/www/resources/admin/classes/common/DBService.php(91): DBService->checkForError()
#1 /home/coral/www/resources/admin/classes/domain/Resource.php(1241): DBService->processQuery('\nSELECT\n  R.res...', 'assoc')
#2 /home/coral/www/resources/export.php(47): Resource->export(Array, 'R.createDate DE...')`

jcuenod:
Thanks! I'll take a look at these.

jcuenod:
I believe that both note types should now be exported but I wasn't quite sure what the other note type was so I haven't tested that (and seen actual content). But I think the things mentioned are fixed.

To be honest, this did make me think I needed to go and do more testing so there may be some lingering bugs here. My time is somewhat limited now, though.

@veggiematts Do exports work for you right now when you hit "new search". There is a pretty egregious problem with the sorting here that I think needs another PR to fix so 10432bb is a bandaid but I'm wondering whether there are other types of escaping that mysql uses (or used to use) - maybe on other platforms...

tristanhoppe:
https://github.com/coral-erm/coral/blob/10432bb85ecb1fc0067599dcd1917d9de43d872e/resources/js/admin.js#L856-L859

The export configuration options don't work in Firefox because Firefox does not support `this.event.` Passing the event in as an argument seems to be the preferred alternative.

jcuenod:
Thanks @tristanhoppe, I'm not able to reproduce this issue. What version of Firefox are you on? (I use Firefox as my primary browser and it's working for me). Although I'm wondering whether this is a config thing in Firefox: `dom.window.event.enabled ` (see `about:config`).

jeffnm:
@jcuenod https://developer.mozilla.org/en-US/docs/Web/API/Window/event says that it's disabled by default in Firefox at the moment and generally a more fragile approach. 

I don't know if we should have users messing with `about:config` flags anyway.

jcuenod:
@jeffnm @tristanhoppe, 4cc5269 should fix the Firefox issue. I've also made failure to commit changes explicit with the `error` callback.

Issue #592, opened 2019-08-13
Import tool: only general subjects can be imported


veggiematts:
Currently, only general subjects can be imported with the import tool. It would be nice to be able to import detailed subjects as well.

PR #593, opened 2019-08-13
Issue 592: Import tool: detailed subjects

veggiematts:
Background: issue #592 : Currently, only general subjects can be imported with the import tool. It would be nice to be able to import detailed subjects as well.

**This patch allows to define detailed subjets columns for the import tool.**

 - Because detailed subjects cannot be attached to a resource without a general subject, detailed subjects are not created: only existing detailed subjects are processed.

 - If the given detailed subject belongs to one or more general subjects, all
   the general subject/detailed subject couples are attached to the resource.

**Expected behavior/test plan:**

 - Import with an existing general subject: the resource is created with the general subject
 - Import with an existing detailed subject: the resource is created with the general/detailed subjects couples
 - Import with a non existing detailed subject: the resource is created, the detailed subject is not created
 - Import with a non existing general subject: the resource is created with the general subject

### DIFF
--- a/resources/admin/classes/common/DatabaseObject.php
+++ b/resources/admin/classes/common/DatabaseObject.php
@@ -243,7 +243,7 @@ class DatabaseObject extends DynamicObject {
                 $query = "INSERT INTO `$this->tableName` SET $set";
                 //echo $query;
                 $this->primaryKey = $this->db->processQuery($query);
-                if ($new) return $this->primaryKey;
+                return $this->primaryKey;
             }
         }
 	}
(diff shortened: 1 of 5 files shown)


poacherone:
Tested, and was unable to import any detailed subjects. Using an existing detailed subject (where the subject is one only one word, i.e. no black space) gives no warning but does not create any detailed subject on the imported record.

Trying to add a detailed subject not existing or one with several words (like my first test subject: "Detailed test") gives a warning:
There was a problem with the database: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 2

A proper warning to the end-user should preferably be added.

streatim:

PR #594, opened 2019-08-26
Dashboards: total cost details

veggiematts:
This development adds a total sum for all the cost details (today there are totals, but one for each cost detail only)

Before:
before
After:
after

Test plan:
 - apply the patch
 - display a yearly costs dasboard with cost details
 - check that you have a new column called "All cost details" that sums the cost details of the line.


veggiematts:
Merged in #598

PR #596, opened 2019-08-26
Dashboards: multiselect dashboards

veggiematts:
Adding the ability to choose several options in the dashboards' dropdown lists.

Before:
before
After:
after

Test plan:
  - apply the patch
  - check that you can select multiple values in the dropdown list, using control+click (the standard html multiselect component is used).


veggiematts:
Merged in #598

Issue #365, opened 2018-02-28
Switching module should not open an new windows


PaulPoulain:
I find counter-intuitive & not good from ergonomics usual behavior to have the "change module" link opening in a new windows.
We had a discussion on coral mailing lists, see comments added in a few seconds

PaulPoulain:
Steve wrote:
- There are in fact things that I routinely do in one module that require me to open a separate module, make a change there, and go back to the original module in its own tab/window to finish inputting information. If we defaulted to always open in same browser window, this would result in loss of work. That is the main reason I am opposed to this proposal. This is particularly true when creating a resource record, realizing that an organization record doesn’t exist, creating that organization record (currently, in a separate window/tab), or modifying something in that record, then going back to fill out the rest of the details in the resource record. But there are other examples I could think of where there is potential loss of work. Perhaps I’m alone in approaching the work in this way; who knows, but I don’t think I am.
- As Eric pointed out, it is also helpful to have two copies of the same module open at times so as to be able to compare or copy/paste information.
- Just because a feature change is easy to implement is not persuasive, in my view. I’m particularly leery of changing default CORAL behavior.
- I am aware of usability standards and best practices and concur with what Jeff Mudge wrote about the normal reasons why a new window/tab is needed vs. open in same browser window.
- A seemingly simple change like what has been proposed needs more careful thought and consideration of ramifications, in my view. We’ve not been great at thinking through that carefully at times in the past, which is why I tend to be cautious now.
- Once we have fully implemented CORAL 2.0 (a release candidate version will hopefully be ready soon, followed by a production release), I’d like to engage the broader community in discussing what should be on a development roadmap, and establish priorities for future enhancements.
- If BibLibre or anyone else feels the need to change the default behavior for moving between modules, you can always do so in your local code, but I’m reluctant to rush to a conclusion for the primary codebase at this time.

PaulPoulain:
My answer to Steve comment:
[...]

PR #606, opened 2019-10-02
Issue 365: Opening modules in the same or a new window is configurable.

veggiematts:
Background: issue #365 : Switching module should not open an new windows

In common/configuration.ini:

```
[settings]
open_new_windows = Y/N
```

By default, the links are opened in a new window/tab.

### DIFF
--- a/common/common_directory.php
+++ b/common/common_directory.php
@@ -161,7 +161,11 @@ function create_date_from_js_format($input) {
   $datepicker_format = return_datepicker_date_format();
   $php_format = create_php_date_format_from_js_format($datepicker_format);
   return date_create_from_format($php_format, $input);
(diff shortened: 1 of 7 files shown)


poacherone:
Tested and switching modules through the top menu works as intended. 

One further question is how to handle the "Helpful links" panel to the right of the page? I would argue that these should also respect the configuration. And there are also a few other places (mostly in resources) that link between modules, what about those?

(Locally we have it set today that only links to external resources/pages open in a new window, all other are within the same window.)

But this is a great starting point!

veggiematts:
Rebased on devel branch. Since it does not change the existing behavior when unconfigured, I'm merging this.

Issue #256, opened 2017-07-22
Slowness with keyword searches in Resources


techsvcslib:
The initial list of resources that displays in the Resources module takes 1-3 seconds to load when there are several thousand (in our case, about 7,800) records. Also, keyword searches in the Name field can take a similar amount of time to retrieve results.

queryluke:
@techsvcslib A Sirsi custom mer with 127,000+ records also mentioned slowness on the resource search page. I poked around with the queries and code and here is my analysis.

# Method

In order to troubleshoot this issue, I wrote [a script](https://gist.github.com/queryluke/93a3dfedc5be49bc1ee29d0ffd13209e) which seeds 250,000 dummy records in a local MySQL db.

# Baseline

Here is the base Resource query:

```sql
SELECT
	R.resourceID,
	R.titleText,
	GROUP_CONCAT(DISTINCT AT.shortName SEPARATOR ' / ') as acquisitionType,
	R.createLoginID,
	CU.firstName,
	CU.lastName,
	R.createDate,
	S.shortName status,
	GROUP_CONCAT(DISTINCT A.shortName, I.isbnOrIssn ORDER BY A.shortName DESC SEPARATOR ' ') aliases
FROM Resource R
LEFT JOIN ResourceAcquisition RA ON R.resourceID = RA.resourceID
LEFT JOIN Alias A ON R.resourceID = A.resourceID
LEFT JOIN ResourceOrganizationLink ROL ON R.resourceID = ROL.resourceID
LEFT JOIN coral_organizations.Organization O ON O.organizationID = ROL.organizationID
LEFT JOIN coral_organizations.Alias OA ON OA.organizationID = ROL.organizationID
LEFT JOIN ResourceRelationship RRC ON RRC.relatedResourceID = R.resourceID
LEFT JOIN ResourceRelationship RRP ON RRP.resourceID = R.resourceID
LEFT JOIN ResourceSubject RSUB ON R.resourceID = RSUB.resourceID
LEFT JOIN Resource RC ON RC.resourceID = RRC.resourceID
LEFT JOIN Resource RP ON RP.resourceID = RRP.relatedResourceID
LEFT JOIN GeneralDetailSubjectLink GDLINK ON RSUB.generalDetailSubjectLinkID = GDLINK.generalDetailSubjectLinkID
LEFT JOIN AcquisitionType AT ON RA.acquisitionTypeID = AT.acquisitionTypeID
[...]

PR #612, opened 2019-11-10
SirsiDynix: Improved selective joins

queryluke:
This issue is related to some Sirsi customers experiencing long resource search load times when they have 100,000 + resources in their system. It is related to #256, but uses a much simpler fix than the one I proposed there.

As explained in #256, the root cause is the ORDER BY statement. But this can be overridden in the [configuration file](https://github.com/coral-erm/coral/blob/master/resources/admin/configuration_sample.ini#L18) by setting default sort to an empty string.

However, after doing this we still saw load times upwards of 30 seconds because the resource index page makes 2 queries per page load. One to [get the resources](https://github.com/coral-erm/coral/blob/master/resources/ajax_htmldata/getSearchResources.php#L35), and another to get the [search result count](https://github.com/coral-erm/coral/blob/master/resources/ajax_htmldata/getSearchResources.php#L24). 

Both queries do a number of left joins, which was additionally adding to the slow load times. The existing code does some selective joining, but there are a number of ["default" joins](https://github.com/coral-erm/coral/commit/c1fc8501dd7ab17303cf18b4b5ede121cd6ae95b#diff-8a0d6c3284dcd145d52d956effda2e8dL843) which aren't required (specifically for the initial page load and searchCount queries).

This is a different approach to the selective join process, which creates an [array of potential joins](https://github.com/coral-erm/coral/commit/c1fc8501dd7ab17303cf18b4b5ede121cd6ae95b#diff-8a0d6c3284dcd145d52d956effda2e8dR771). Each item in the array is a key value pair:

__key__ the SQL alias for the joining table
__value__ and array consisting of:
1. __stmt__: the joining sql statement
2. __requires__ (optional): the table that is required to join the desired table to the resource table.

For example:

The AT (AcquisitionType) table cannot be directly linked to the Resource table. It is linked through the  ResourceAcquisition table. Thus, `"requires" => "RA"` will result in the following SQL

```sql
LEFT JOIN ResourceAcquisition RA ON R.resourceID = RA.resourceID
LEFT JOIN AcquisitionType AT ON RA.acquisitionTypeID = AT.acquisitionTypeID
```

### DIFF
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -529,22 +529,20 @@ class Resource extends DatabaseObject {
 
 			if ($config->settings->organizationsModule == 'Y') {
 				$dbName = $config->settings->organizationsDatabaseName;
-
 				$whereAdd[] = "((UPPER(R.titleText) LIKE " . $nameQueryString . ") OR (UPPER(A.shortName) LIKE " . $nameQueryString . ") OR (UPPER(O.name) LIKE " . $nameQueryString . ") OR (UPPER(OA.name) LIKE " . $nameQueryString . ") OR (UPPER(RP.titleText) LIKE " . $nameQueryString . ") OR (UPPER(RC.titleText) LIKE " . $nameQueryString . ") OR (UPPER(RA.recordSetIdentifier) LIKE " . $nameQueryString . "))";
 
 			}else{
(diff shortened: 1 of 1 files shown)


andyp-uk:
Tested in house and in production on a customer system. Looks good.

PR #613, opened 2019-11-14
License export

tristanhoppe:
**Purpose**
This is an adaption of jcuenod's resource export feature which is being applied to the license module.

**Issue**
There may be some files or modifications copied from jcuenod's version that are not needed in this implementation.


streatim:

PR #619, opened 2019-12-06
SirsiDynix: Additional Datepicker fixes

queryluke:
This is an extension of #581 and #582 

Prior to this fix, dates within a form are formatted by the datepicker plugin. Thus, in the config file, if the format were `dd/mm/yyyy`, the date __January 2nd, 2019__ would be formatted _and submitted_ as the string `02/01/2019`.

But the php scripts that parse the `$_POST` variables are using [`strtotime()`](https://www.php.net/manual/en/function.strtotime.php), which "..expects to be given a string containing an English date format". Further digging into php docs showed that strtotime [only works with American date formatting](https://www.php.net/manual/en/datetime.formats.date.php) when using forward slashes, and has very specific formats for dashes.

Thus the above string of `02/01/2019` will always be parsed as __February 1st, 2019__ using `strtotime()`

Here, I've written the necessary functions to use [`date_create_from_format()`](https://www.php.net/manual/en/datetime.createfromformat.php), which, instead, lets you pass a specific format to the parser, so the parse knows exactly where days and months exist in the string.

## Testing
1. In `common/configuration.ini` Change the `datepicker_date_format` and `date_format` settings to something else (e.g. `dd-mm-yyyy` and `%d-%m-%Y`, respectively).
2. Create a new resource
3. Edit the resource's order (or add a new one) and change the subscription start-end dates.
4. The dates entered in the form should persist after the order information is submitted (prior to this fix, a data such as `20-12-2019` would have resulted in a 1969 date because strtotime cannot parse that string.

### DIFF
--- a/common/common_directory.php
+++ b/common/common_directory.php
@@ -98,6 +98,38 @@ function format_date($mysqlDate) {
 
 }
 
+function create_php_date_format_from_js_format($input_string) {
+
+  // Note the js format strings are specific to the datepicker plugin
+  $js_formats = array('/yyyy/','/yy/','/mmmm/','/mmm/','/mm/','/dd/');
+  // php format strings https://www.php.net/manual/en/function.date.php
+  $php_formats = array('Y','y','F','M','m','d');
+
+  return preg_replace($js_formats, $php_formats, $input_string);
+}
+
+function create_date_from_js_format($input) {
+  /*
+   * see https://andy-carter.com/blog/php-date-formats for overview of different php date formatters
+   * Coral utilizes strftime() and strtotime(), but strtotime expects dates to be formatted in US English
+   *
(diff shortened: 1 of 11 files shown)


andyp-uk:
Fix looks good. We are now able to use dd/mm/yyyy format for display and input, if configured.

In testing, we notice the date format used by CORAL does not respect the values configured in the common/configuration.ini file in the following places. However, these all use the international format (yyyy-mm-dd) and therefore are not likely to be misinterpreted. I didn't want to change them without further discussion. 

- Resources module — Resource record — Order drop-down list (appears when you have two or more orders)
- Resources module — Resource record — Issues tab — under "view open issues"
- Usage Statistics module — SUSHI page — on the dialog box after clicking "run now"
- Management module — Home page — Last Document Revision

Issue #475, opened 2018-08-31
Issue with document upload Resources Module


rtatterson:
There appears to be an intermittent issue getting documents to upload in Resources. At times it requires several attempts to get them to attach.

veggiematts:
I can confirm this behavior.

sekjal:
Confirmed in Licensing module, as well

Issue #458, opened 2018-07-19
Coral silently fails on too large file uploads


veggiematts:
When adding an attachment to a resource that exceeds php's post_max_size limit, coral will still display "{filename} successfully uploaded." and the attachment will be created.

This will lead the user to believe that the file has been successfully uploaded when it's not.

We should display an appropriate error message in this case.

PR #620, opened 2019-12-11
Upgrade file uploads

veggiematts:
This PR fixes issues #475 and #458.

    This patch adds meaningful error messages when an upload fails for any reason.
    
    This is more a rewrite than a bugfix: it replaces the use of AjaxUpload
    (code from 2008, apparently not maintained anymore) with a standard JQuery
    ajax upload.
    
    It handles both PHP upload errors
    (see https://www.php.net/manual/en/features.file-upload.errors.php)
    and webserver errors ("413 Entity Too Large" for example)
    
    In order to have standard JQuery ajax upload, jquery needs to be upgraded.
    
    This patch upgrades jquery from 1.4.4 to 1.8.0
    
    The original datepicker was not compatible with jquery 1.8.0, so it has been
    updated with a fork of the existing one:
    https://github.com/michaelulguim/jQuery-Datepicker--Kevin-Luck-
    It's not a new datepicker, it's an update of the existing datepicker.
    
    Some cleanup has also been done:
     - ajaxupload plugin has been removed
     - unused older jquery versions were remove

The following file uploads have been updated:
 - Attachments to resources
 - Attachments and documents to licenses
 - Attachments and documents in management.

Test plan:
 - Apply the patch
 - Try uploading a file that exceeds the maximum size allowed by your webserver or php configuration
 - Check that you got an error message
 - Check that datepickers still work.

Capture d’écran_2019-12-11_16-49-26


poacherone:
Tested, and works very well. I seem to only get 500 errors when the file is too large, but I'm assuming this is because of my test set-up (Apache I would guess?). One of my documents folders was not writeable, and that error displayed correctly.

veggiematts:
Thanks for testing, @poacherone 
Merging this.

Issue #601, opened 2019-09-18
Behaviour of browser back-button and Coral


poacherone:
In the resource module, if you do a search for a resource (or browse to another pagination, use one of the filtering options etc.), click through to a resource, and then use the browsers back-button the behaviour is today that the values from the search/browsing/filtering is gone and all resources display. The expected behaviour from staff here would be for their searches (etc.) to be kept.

Various browsers handles this differently though. Chrome for example will keep a search or filtering, but not pagination, while Firefox retains none of the options.

If it is not possible to provide the same behaviour in all browsers, then perhaps a link back to the search results can be displayed on the resource somewhere (a common option used in other systems).

ehartnett:
I'd be in favor of this functionality being added. The Licensing module retains your search when you go into a record and then back out. Could the same code be used in Resources?

veggiematts:
Actually this feature seems to exist already for the resource module but is somehow broken. I'll try to come up with a fix.

PR #621, opened 2020-01-14
Issue 601: Keep search parameters in resource module

veggiematts:
Background, see Issue #601 
 
This patch fixes the previously existing behavior that was broken:

 When going back from a resource detail to the index page using the Home menu
 link or using the back button of the browser, the search parameters and results
 are kept.

This patch is not very pretty code-wise, so if anyone has a better solution, they're welcome to make another patch.
If not, this can be a quickfix for the time being.

### DIFF
--- a/common/common_directory.php
+++ b/common/common_directory.php
@@ -161,7 +161,11 @@ function create_date_from_js_format($input) {
   $datepicker_format = return_datepicker_date_format();
   $php_format = create_php_date_format_from_js_format($datepicker_format);
   return date_create_from_format($php_format, $input);
+}
 
+function getTarget() {
+    $config = new Configuration();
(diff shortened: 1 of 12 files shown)


sbouis:
I have just tested it on Chrome. It works for almost all the search exept for "fund" and "relationship".
For those two, when you click on "home" or use the back button, you retrieve all the resources.

veggiematts:
Thanks for the feedback @sbouis !
The branch has been rebased to avoir conflicts, and both fund and relationship filters are now kept.

veggiematts:
I'm merging this.

poacherone:
Tested after the rebase, on both Chrome and Firefox, and works as intended! Great work!

Issue #623, opened 2020-01-28
Add the possibilty to find the "adults" without child in the resource module


sbouis:
In the resource module main page, there's the possibility to retrieve parent resources and the child resources with the "relationship" filter. It would be great if we could also retrieve resources that are neither a child nor a parent, let say "adult" resources.

PR #624, opened 2020-01-30
Issue 623: Search resources with no relationship

veggiematts:
Background: see issue #623 

Currently, resources can be filtered with the following relationship criteria:

 - **All**: all resources are displayed (child, parent, or no relationship)
 - **Parent**: only parent resources are displayed
 - **Child**: only child resources are displayed

This patch adds the following criteria:

 - **None**: only resources that are neither parent nor child are displayed
   (ie: only resources with no relationship)

Test plan:
 - Perform a search on the resources index page, with 'None' in the Relationship
   dropdown list.
 - Check that only the resources with no relationship are displayed.

### DIFF
--- a/resources/admin/classes/domain/Resource.php
+++ b/resources/admin/classes/domain/Resource.php
@@ -555,9 +555,12 @@ class Resource extends DatabaseObject {
 
 
 		if ($search['parent'] != null) {
-			$parentadd = "(" . $search['parent'] . ".relationshipTypeID = 1";
-			$parentadd .= ")";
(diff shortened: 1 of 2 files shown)


sbouis:
Ok, it works as intented.
The new value "none" is displayed in the drop-down menu "relationship".
I searched for "parent", "child" and "none" and I have the resources corresponding for each search.

veggiematts:
Thanks for testing @sbouis 
Small change, unlikely to break anything, I'm merging this.

PR #631, opened 2020-02-21
Add common classes to the reports module

veggiematts:
The reports module is currently broken:

PHP Fatal error:  Uncaught Error: Class 'Configuration' not found in /home/coral/www/common/common_directory.php:78

This patch adds the common classes also used by the other modules and fixes the problem.
This patch should be merged before release.


poacherone:
Tested this. The module is broken without the patch, and works with it applied.

Issue #636, opened 2020-02-25
"Total" field is missing in dashboard_yearly_costs export


oliviercrouzet:
needs to add this (line 66) :
```
+    $columnHeaders[] = _("Total");
     echo array_to_csv_row($columnHeaders);
```

and this (line 93) :
```
+        $dashboardValues[] = $result['costDetailsSum'];
         echo array_to_csv_row($dashboardValues);
```

PR #637, opened 2020-02-25
Issue 636: Add missing "All detail costs" column in yearly costs dashboards exports

veggiematts:
Background: Issue #636  

Test plan:

 - Check that the "All detail costs" column is currently missing from yearly
   costs CSV exports (although displayed in the yearly costs page)

 - Apply the patch

 - Check that the column is now in the CSV exports and the values are correct.


poacherone:
Tested, works as expected!

Issue #638, opened 2020-02-26
All cost details show wrong sum in dashboard


poacherone:
On the development branch, the column "All cost details" doesn't show the correct total sum, but instead multiplies it (double or tripling the correct sum). 

This seems related to #603, but here applies to "All cost details" (and not necessarily to multiple subjects, I've found examples of tripled sums with only one subject) . #576 changes a lot of the code on the sums of the dashboard. Each cost detail has an DISTINCT added before the SUM, and I think the sum of all cost details need a DISTINCT as well to work as intended. A quick local fix seems to have resolved the issue.

PR #639, opened 2020-02-27
Issue 638: Fix yearly costs dashboards sums for resources with multiple subjects

veggiematts:
Background: issue #638 

Test plan:

 - Find a resource with several subjects attached.
 - Check that the 'All cost details' value is wrongly
   multiplied by the number of attached subjects.
 - Apply the patch.
 - Check that the 'All cost details' value is now correct.


poacherone:
Tested, works great! Also checked both single years and multiple - correct sums displayed in both cases.

PR #641, opened 2020-03-09
2020.02 installer components

jeffnm:
Changes to to the unified installer scripts to add the new version and the one new DB change. 

### Fresh Install

Clone the repository and walk through a fresh installation. Make sure the common/configuration.ini version reaches the right number and that the organizations table has the ilsID column. 

### Update

With a copy of an existing installation, copy the new code into the directory. You should be prompted to update the existing installation and asked for a DB username and password that can update the database. Verify that the new version is installed by checking the common/configuration.ini and the organizations table's ilsID column exists.

### Misc

The footer hasn't been updated yet (that happens on Master branch once this the release is merged, but before the release is published) so it says Development still. 

I have tested the fresh install and got a message about one of the 3.0.1 SQL lines, but after refreshing, the installer completed and things looked OK. 

I have not yet tested the update path, but I wanted to get this available for others who are doing testing.

PR #644, opened 2020-03-26
Revert "2020.02 installer components"

xsong9:
Reverts coral-erm/coral#641


jeffnm:
I'm going to close this pull request, because it's reverting the installer merge and the steering committee has agreed that we need to keep those changes and proceed with the release.

PR #650, opened 2020-04-13
Helpful links respects new tab/window confguration

poacherone:
The helpful links in the right-hand panel should also respect the configuration of whether new tabs/windows should be opened or not (set in common/configuration.ini). This PR adds this functionality.

The one exception is the "Print view" link, which still opens a new tab/window, since this is a common UX for print functionality. But this can be discussed!


poacherone:
I messed up and based my work on top of master. Closed this PR and made a new one, on top of development: [(https://github.com/coral-erm/coral/pull/651)]

PR #654, opened 2020-05-10
Add qualifiers to license Expressions tab for "view only" user

andyp-uk:
A SirsiDynix customer reported that it is not possible to view the license expression qualifiers on the Expressions tab of the license record if the user has the "view only" privilege. This seems unintentional since the same user can view the qualifier on the Expression Comparison page. Moreover the tool tip for the user admin states "View only users can view all license information, including the license pdf."

This change adds the qualifier to the Expressions tab for users with the "view only" privilege. The option to edit or remove the expression does not appear.


andyp-uk:
Have tested this and can verify the change. It's a small change so I feel confident to merge it even though it's my own work.

PR #660, opened 2020-07-31
COUNTER 5 Usage Imports

queryluke:
## Feature
Makes the usage module compatible with COUNTER 5 reports. Backwards compatible with COUNTER 4.

## Requirements
Run `usage/install/protected/next/001-01.sql` against your development database.

## Testing
- Create new SUSHI COUNTER 5 services and test with various vendors
- Download various COUNTER 5 reports from vendors and manually upload
- Test with SUSHI COUNTER 4 vendors
- Test with COUNTER 4 reports from vendors


andyp-uk:
Note the Apache user should have full permissions to the usage/counterstore folder where applicable, in common with usage/logs and usage/archive folders.

usage/admin/classes/domain/SushiService.php
```diff
@@ -432,6 +432,18 @@ private function jsonTransfer($reportLayout, $serviceProvider)
...
       foreach($json->Report_Header->Exceptions as $exception) {
+
+        // 1011 means the report was queued, in this case, try again after 1 second
+        // after three retries, exit
+        if ($exception->Code == 1011) {
+          if ($retry == 3) {
+            $this->logStatus("Report was queued for processing, but was not processable after 3 retries.");
+            $this->saveLogAndExit($reportLayout);
+          }
+          sleep(1);
```
andyp-uk, line 443: I think we should back out this commit. Using JSTOR as the only example of a vendor that returns error 1011 that we are aware of, the wait time varies considerably, depending on the number of reports you request and the date range you specify. Even with a very long sleep interval, I don’t get any reports more often than I do. The user can always manually rerun the import from the console.

andyp-uk:
We will need to rename this file to match the version number of the release.
usage/install/protected/3.0.2/001-000.sql

andyp-uk:
@rtatterson has tested these changes and reports no problems.

PR #662, opened 2020-08-26
Fix to downtime logging feature

andyp-uk:
These changes address several issues with downtime logging:

- When entering downtime, all dates were incorrectly recorded as being in 1969. This was caused by a fix to allow configuration of date format used for the date picker tool. 
- Unable to enter downtime in the Organizations module (due to a typo on startDate).
- Unable to export downtime from the Resources module.

I've also made some very minor style changes to make it easier to see where one downtime definition ends and the next one starts in cases where there are several listed.


andyp-uk:
Tested in-house and fix looks good, therefore merging.

PR #671, opened 2020-10-15
Contacts appear blank when reporting issues if name is empty

andyp-uk:
Reported by a SirsiDynix customer: when reporting a new issue for a resource or organization, contacts are shown as blank values when the contact has no name. It would be preferable to show the email address when there is no name.

These are relatively simple changes that ought not adversely affect anyone who is content with the current functionality.

It's perhaps a moot point whether it's good practice to enter contacts without a name although I note the name field is not strictly required.


andyp-uk:
Tested in-house successfully.

PR #691, opened 2021-11-08
PHP 8 Compatibility changes

aniqaamir:


andyp-uk (changes requested):
I think you intended to make changes to sirsidynix/coral:master and not coral-erm/coral:master. In general we do not push changes directly to coral-erm/coral:master. Instead, we ask the community to update coral-erm/coral:development.

Issue #695, opened 2022-01-24
Reports.php: Downloaded XLS file doesn't have the correct file suffix


PerplexedTheta:
Hi all,

First of all, this is my first submission to this community's issue tracker. I have attempted to adhere to generic bug submission conventions, however, if there is a problem with the place or methodology I have used in submitting this bug, please do let me know.

On to business: we have been experiencing an issue on our Coral ERM sites around the reports module. Namely, when you attempt to export a report as an XLS file, the download delivers a file that would appear to be rubbish data. Upon further inspection, I've found Excel capable of opening the file simply by adding the 'XSL' file extension.

Steps to reproduce:

1. Navigate to /reports
2. Select a report which contains some data
3. On the results page, select the Excel icon
4. The file is downloaded successfully, but without the file extension

I have appended a patch to this ticket, which, I'm sure you'll agree, is a simplistic fix to the issue. If further work is desired, or necessary, to get this patch into the product, please do let me know, and I will gladly assist! 

[coral-report-xls-export-hotfix.patch.txt](https://github.com/coral-erm/coral/files/7925107/coral-report-xls-export-hotfix.patch.txt)

andyp-uk:
Please could you create a pull request with the code changes and link it to this issue? The pull request should merge changes into the development branch.

PerplexedTheta:
Hi Andy,

Thanks for the comment. I have created a pull request (new territory for me), which can be viewed [here](https://github.com/coral-erm/coral/pull/696).

Please let me know if I can assist further

streatim:
Per a conversation in PR #697, @gmcharlt wrote  
> Better long term options might be to provide a CSV export option and/or a library to generate true XLS or XSLX output.

I'm putting this note here to indicate what the future release recommendation would be; likely a .csv export rather than XLS.

PR #697, opened 2022-01-24
Reports: Downloaded XLS file now has the correct file suffix

PerplexedTheta:
Please see [issue 695](https://github.com/coral-erm/coral/issues/695)

### DIFF
--- a/reports/report.php
+++ b/reports/report.php
@@ -89,7 +89,7 @@ if ($outputType === 'print') {
     header("Content-Type: application/download");
     header("Content-Transfer-Encoding: binary");
     header("Content-type: application/vnd.ms-excel;");
-    header("Content-Disposition: attachment; filename='" . strtr($report->name, ' ', '_') . "'");
+    header("Content-Disposition: attachment; filename=" . strtr($report->name, ' ', '_') . ".xls");
 ?>
     <html>
         <head>


PerplexedTheta:
Hi,

Any update on this request?

Thanks,
Jake

andyp-uk:
What you've done looks fine. Someone will need to test then merge it. If nobody else gets to it first, I will very likely do this myself before the next release.

andyp-uk:
Having a look at the issue myself, I see the file generated isn't actually an XLS file. Adding an XLS file suffix makes it easier for users to open the file with Excel, but there is still a warning that the file could be corrupted or unsafe. 

Other members of the CORAL steering committe might want to give their opinion before this is merged.

gmcharlt:
While looking at this, I ran across the following: back in 2016, Microsoft released a [security update](https://support.microsoft.com/en-us/topic/ms16-088-description-of-the-security-update-for-excel-2013-july-12-2016-5dd955cd-4464-47bf-b7e3-5ab2b08e5953) that made Excel refuse to load HTML files that were given an .xls extension at all (unless the website was added to a trusted list). Microsoft [subsequently changed the behavior](https://support.microsoft.com/en-us/topic/august-9-2016-update-for-excel-2016-kb3115438-8c93049f-a1b7-8919-6675-ba8491ef4567) so that such files would instead be opened in "protected view". Consequently, there are theoretically some users who may not be able to open such files in Excel.

In any event, I do get a warning from Excel that the file may be corrupted. Better long term options might be to provide a CSV export option and/or a library to generate true XLS or XSLX output.

PR #678, opened 2021-03-05
Update externalLoginSubmitForm.js

mirkh:
In Usage module: The modal "getLoginForm" will not open. 

Usage module Home, click on a Platform entry, click on the tab Logins, click "add new login" There will be a waiting image, but the modal is not opening.

The problem was two unclosed parentheses in the javascript file.

PR #699, opened 2022-03-21
Usage module getLoginForm will not open

andyp-uk:
Per @mirkh 

> In Usage module: The modal "getLoginForm" will not open. 
> 
> Usage module Home, click on a Platform entry, click on the tab Logins, click "add new login" There will be a waiting image, but the modal is not opening.
> 
> The problem was two unclosed parentheses in the javascript file.


andyp-uk:
Previously reviewed at https://github.com/coral-erm/coral/pull/678

PR #701, opened 2022-04-27
Add platform parameter to SUSHI Services in Usage Module

kmcarnes:


gmcharlt:
I have tested the change and confirm that

- it's properly saving the platform in the SUSHI service definition
- it's allowing me to successfully fetch reports from Loeb Classical Library (which requires a platform string)

PR #710, opened 2022-10-12
Modern jquery update

spaceisntsyntax:
This PR represents the changes necessary to make use of jquery 3.6.0 from (currently) the Google CDN.  It also replaces the thickbox modal widget with the jquery-ui dialog widget, which is both maintained and does not require the use of inline iframes.


bbjonn:
I think you might have missed a couple of references -  running a grep from the home directory. I'm not sure these are actually doing anything but they get flagged in a pen test.
index.php:      <script type="text/javascript" src="js/plugins/jquery.js"></script>
management/documents/index.php: <script type="text/javascript" src="js/plugins/jquery.js"></script>

andyp-uk:
Changes are included under #759 .

PR #711, opened 2022-10-12
Modern jquery update with PHP8 support from SirsiDynix

spaceisntsyntax:
This PR contains all of the changes from issue #710 as well as the commit from SirsiDynix that updates PHP support to version 8.

Minor merge conflicts between the PHP commit and the development branch were handled by keeping the original branch's code, while the jquery commits were preferred for the later merges.  It may be the case that the PHP side of conflicts should be chosen instead, but scanning the conflicts suggested they were primarily transient and would be superseded by the jquery branch.


andyp-uk:
To be addressed by #759 instead.

PR #749, opened 2024-03-01
[747] Updates for jquery datepicker plugin and javascript includes cleanup

### LINKED ISSUE #747: Error with datepicker for at least one month
The datepicker is having trouble with November 2024:

30-days-hath-november

jsavell:
This PR:

- Resolves #747 
- Removes redundant loading of javascripts from resource.php that have already been loaded in [header.php](https://github.com/coral-erm/coral/blob/ERM-63-modern-jquery-update-with-php8/resources/templates/header.php#L54)
- Restores the i18n keys to the datepicker plugin
- Updates the datepicker plugin to use the latest version from the original author: http://2008.kelvinluck.com/assets/jquery/datePicker/v2/demo/scripts/jquery.datePicker.js
- Attempts some code formatting cleanup

The latest, original version of the plugin appears to include all the modern jquery updates that were provided by the forked version along with additional improvements and fixes. I haven't found any issues testing these changes locally.

### DIFF
--- a/js/plugins/jquery.datePicker-patched-for-i18n.js
+++ b/js/plugins/jquery.datePicker-patched-for-i18n.js
@@ -1,22 +1,17 @@
 /**
- * Copyright (c) 2007 Kelvin Luck (http://www.kelvinluck.com/)
- * Dual licensed under the MIT (http://www.opensource.org/licenses/mit-license.php) 
+ * Copyright (c) 2008 Kelvin Luck (http://www.kelvinluck.com/)
+ * Dual licensed under the MIT (http://www.opensource.org/licenses/mit-license.php)
  * and GPL (http://www.opensource.org/licenses/gpl-license.php) licenses.
- *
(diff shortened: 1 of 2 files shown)


andyp-uk:
@jsavell thanks for this. How much testing do we need to do? Can you please give us some guidance? Will these changes impact other modules apart from Resources?

jsavell:
Hi, this date picker plugin is shared across multiple/all modules, so the fix impacts/benefits all usages of the '.date-pick' class hook.

For testing, just verifying that the date picker pop ups still appears should be enough. If your installation wasn't affected by the DST bug, there shouldn't be any visible differences before/after.

In applying this PR to this 'ERM-63-modern-jquery-update-with-php8', I did find a case where the date picker pop up wasn't applying properly: https://github.com/coral-erm/coral/blob/ERM-63-modern-jquery-update-with-php8/management/ajax_forms.php#L121

I tested this issue on the parent 'ERM-63-modern-jquery-update-with-php8' without my update, and the problem was already present. I also tested on the Coral Demo site and the problem was *not* present.

PR #752, opened 2024-03-20
Fixed characters encoding issue ERM-116

sumbria:
ERM-116

PR #754, opened 2024-03-21
Fixed utf8 encoding issue ERM-116

sumbria:


andyp-uk:
This change is intended to fix an issue where replacement character appears instead of the correct accented characters on the Calendar page of the Licensing module. This appears to be happening when either the authorized site or resource type contains accented characters. This supersedes #752

PR #757, opened 2024-03-27
ERM-117 

sumbria:
- Added in-progress statuses option in admin
- Get license in progress dynamically


andyp-uk:
This PR addresses an issue reported by a French Canadian user whereby the License In Progress page doesn't work after changing some of the delivered license status values to use French instead. These changes add a configuration option to control which license statuses are qualifying.

PR #758, opened 2024-04-03
Export issues & resolve button fix

sumbria:


andyp-uk:
This PR fixes the following.

- 500 error when exporting issues or downtime from Resources module in cases where the report is empty.
- Submit button doesn't work when attempting to resolve downtime.

PR #759, opened 2024-04-04
Modern jQuery update with support for PHP 8

andyp-uk:
This PR addresses the following.

- Updates PHP support to version 8.
- Update to use jQuery 3.6.0 from (currently) the Google CDN. It also replaces the thickbox modal widget with the jquery-ui dialog widget, which is both maintained and does not require the use of inline iframes.
- Various bug fixes reported during testing of the above changes.


spaceisntsyntax (approved):
Recent changes all look good, and prior commits have been well tested to date.  Let's go!

PR #823, opened 2025-07-25
V2025.04.05 (#822)

streatim:
* Update getProductDetails.php

Removes Resource Note ID to resolve Issue 817.

* Adds Toggleable Notice to Usage Stats Main Page

Adds a toggleable notice to the main page of the Usage Stats module, letting folks know that development is seeking volunteers. Also styles the page and sets up javascript for the toggle button.

* Half Resolves Workflow Issue

Updates adding and removing members of a group in the workflow tab. Half resolves Issue 771 (which also identified an issue with the Workflow Steps form).

* 2024.04.05 Updates

Updates the installer for v2024.04.05.

### DIFF
--- a/auth/install/auth.php
+++ b/auth/install/auth.php
@@ -448,7 +448,17 @@ function register_auth_provider()
 					return $return;
 				}
 			];
-        
+
+		case "2025.04.05":
+			return [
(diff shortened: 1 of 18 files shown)


rtatterson (approved):
These look good.

PR #836, opened 2025-11-21
v2025.10.01

streatim:


rtatterson (approved):
Everything checked out okay, so approving merge.

